PR #24, opened 2017-06-15
use namespace

strengejacke:


JasperHG90:
I use @importFrom sjlabelled in the readSurvey() function. Is it not working?

strengejacke:
Yes, it works, however, inside package functions, you have to qualify the function call with `<pkg>::<function>` when referring to other functions. I think this is obligatory since R 3.x.

JasperHG90:
Ah, thanks for this!

Love your work btw :-)

Issue #41, opened 2017-07-11
Possible Encoding Issue with Windows 10


shiyaoliu:
Hi Jasper,

I tried to getSurvey from my qualtrics, but got the following error message. My platform is Windows 10 Professional Simplified Chinese. The same code works under Ubuntu 16.04 LTS. So I guess it's probably an encoding issue.

Thanks.

Best,
Sean

> mysurvey.2 <- getSurvey(surveyID = surveys$id[2])
  |======================================================================================| 100%
Error in read.table(file = file, header = header, sep = sep, quote = quote,  : 
  no lines available in input
In addition: Warning messages:
1: In readLines(file, skip) :
  invalid input found on input connection '.../China Survey.csv'
2: In readLines(file, skip) :
  incomplete final line found on '.../China Survey.csv'

JasperHG90:
This is the same error as #36 and #38. Can you install version 2.1 from the master branch and tell me if that solves the problem? Thanks!

shiyaoliu:
Thanks for the quick response. I downloaded v2.1 and tried it out.

This partially solved the problem. When the questions in qualtrics are in English, the code works now. 

However, when questions and responses contain East Asian languages, say Chinese, there is another error message. This is probably caused by the operation system encoding compatability I guess. Based on my previous experience dealing with East Asian languages in R, it seems that R in Windows needs ANSI encoding, while R in Linux supports Unicode. 

mysurvey <- getSurvey(surveyID = surveys$id[7],
+                       save_dir=getwd())
  |======================================================================================| 100%
Argument `label` must be of same length as numbers of columns in `x`.
 Show Traceback
 
 Rerun with Debug
 Error in as.POSIXlt.character(x, tz, ...) : 
  character string is not in a standard unambiguous format

JasperHG90:
The getSurvey() function takes a 'fileEncoding' argument (from the read.csv() function in the utils package). Could you see if one of the encodings provided works for your survey? Thanks!

JasperHG90:
Also: try to run it without inferring data types!

JasperHG90:
Hi Shiyaoliu,

Did you get it to work?

Thanks!

J.

shiyaoliu:
Sorry I have been occupied with some other projects. Will get back to you as soon as I try it out tomorrow.

Thanks.

JasperHG90:
Thanks!

JasperHG90:
Hi Shiyaoliu,

[...]

Issue #38, opened 2017-07-04
Unknown Warning


Delkine:
I'm receiving the following warning when I us qualtRics to download a survey.  
"Warning message:
In scan(file = file, what = what, sep = sep, quote = quote, dec = dec,  :
  invalid input found on input connection 'C:/Users/del/AppData/Local/Temp/RtmpQZ7nvM/Mindshare English v21.csv'"

Can you point me some help with this?  I'm also not getting all the respondent surveys in the download.

Del

JasperHG90:
Hi Del,

Thanks for your message. I think this error is the same as #36. 

As you can see from that thread, it's quite difficult for me a) reproduce this error on my end and b) debug it at a distance.

Would you be open to skyping and sharing your screen with me so I can see the exact steps you're taking? It would speed up the debugging process.

If not, then we'll try to do this at a distance.

Thanks for considering!

J.

Delkine:
I'm happy to do  Skype with you but not today.  I'm in California and by the time  I get home it will be too late.   Can you do it tomorrow?
Del

Sent from my Verizon, Samsung Galaxy smartphone
-------- Original message --------From: Jasper Ginn <notifications@github.com> Date: 7/4/17  10:29 AM  (GMT-08:00) To: JasperHG90/qualtRics <qualtRics@noreply.github.com> Cc: Delkine <dnebeker@cox.net>, Author <author@noreply.github.com> Subject: Re: [JasperHG90/qualtRics] Unknown Warning (#38) 
Hi Del,
Thanks for your message. I think this error is the same as #36.
As you can see from that thread, it's quite difficult for me a) reproduce this error on my end and b) debug it at a distance.
Would you be open to skyping and sharing your screen with me so I can see the exact steps you're taking? It would speed up the debugging process.
If not, then we'll try to do this at a distance.
Thanks for considering!
J.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  




[...]

Issue #36, opened 2017-06-27
Data imported to R is not current


cmbohn:
When importing survey data into R, is there a delay between what is current in the Qualtrics system and what is being downloaded?

Is it possible that the data is having problems importing? I enter the following into R:
` loyperc <- getSurvey("SV_5nIBD5iRalXmdNz", useLabels=FALSE, useLocalTime=TRUE, dateWarning=FALSE, convertStandardColumns=TRUE, force_request=TRUE) ` 
I get the following warning:
 `Warning message: In scan(file = file, what = what, sep = sep, quote = quote, dec = dec, : invalid input found on input connection 'C:/Users/cb37543/AppData/Local/Temp/RtmpeEgLjr/Loyalty perceptions (Jamie).csv' ` 
I can only import 88 observations, when there is actually much more than this.

cmbohn:
Is it possible that the data is having problems importing? I enter the following into R:
`loyperc <- getSurvey("SV_5nIBD5iRalXmdNz", useLabels=FALSE, useLocalTime=TRUE, dateWarning=FALSE, convertStandardColumns=TRUE, force_request=TRUE)`
I get the following warning:
`Warning message:
In scan(file = file, what = what, sep = sep, quote = quote, dec = dec,  :
  invalid input found on input connection 'C:/Users/cb37543/AppData/Local/Temp/RtmpeEgLjr/Loyalty perceptions (Jamie).csv'`
I can only import 88 observations, when there is actually much more than this.

JasperHG90:
Thanks for your comment.

> When importing survey data into R, is there a delay between what is current in the Qualtrics system and what is being downloaded?

I'm not aware of such a delay. Essentially, you're just making a call to the Qualtrics API in which you request a csv file of the response exports. 

> Is it possible that the data is having problems importing?

Yes, that is possible. (see e.g. #13). In this case, I'm guessing there are special characters in your csv file. I googled that error and found the following:

- [Invalid Input causes read.csv to cut off data](https://stackoverflow.com/questions/19610966/invalid-input-causes-read-csv-to-cut-off-data)
- [Invalid multibyte string in read.csv](https://stackoverflow.com/questions/14363085/invalid-multibyte-string-in-read-csv?rq=1)
- [Reading in data with unusual characters](https://stackoverflow.com/questions/19301265/reading-in-data-with-unusual-characters)

[...]

PR #43, opened 2017-07-17
Fix a bug where one line too much is skipped in readSurvey.R.

EruIluvatar:
Hi Jasper,
Stefan and Adrian here.

We noticed that always the first row of data is missing when using `readSurvey`.
This PR reverts a change (skip = 2 -> 3) that you probably introduced when adapting our original code to the package.

Also the `importids` are never used after creating it. Which is why we remove it.

Thanks for considering this PR!


EruIluvatar:
I just realized that qualtrics changed their export formats since the beginning of the year.
We have worked with the now-called "legacy" format, it seems you adapted it to the now "standard" format when integrating it?

My suggestion is that we introduce another paramater `legacy` which defaults to `FALSE`.
When set to `TRUE`, we want to only skip 2 lines.

How do you feel about that?
Needless to say that this PR is not ready for merge from my side. 

---
Attached: sample of standard and legacy format. Feel free to use them directly in the test.
[readSurveyTest_export-samples.zip](https://github.com/JasperHG90/qualtRics/files/1155029/readSurveyTest_export-samples.zip)

EruIluvatar:
Sorry for the amount of changes, most lines are because of refactoring tabstop 4 to 2.
Makes the diffs a bit messy...

JasperHG90:
Hi Adrian, Stefan,

Thanks for your comments! 

One question: is it absolutely necessary to use UTF encoding? Could create problems on Windows machines. See issues #36, #41 and #38.

J

EruIluvatar:
Thanks for your feedback Jasper.

Just to make clear, the UTF-8-BOM encoding is only enabled if we want to read a legacy csv export.

The problem is that the [sample file using the legacy export](https://raw.githubusercontent.com/EruIluvatar/qualtRics/b8d9c7ec7621dea34281113aa3c6bb59e84e390e/tests/testthat/files/sample_legacy.csv) always has UTF-8 with byte order mark encoding as obtained from qualtrics.com.
The byte order mark is another byte at the start of the file, causing a corrupted first entry (V1 turns into something like ï»¿V1).
[See more at stackoverflow](https://stackoverflow.com/questions/3255993/how-do-i-remove-%c3%af-from-the-beginning-of-a-file#3256014). This was the reason we put "UTF-8-BOM" in there in the first place, because an empty parameter, "latin1" or "UTF-8" would corrupt our first column name.

Now we observed that while the legacy exports have the UTF-8 with BOM, the new (standard) exports have regular UTF-8 encoding _without_ BOM. That's why the encoding is only set when using readSurvey with `legacyFormat = TRUE`.

For what it's worth we tested this on Windows 10 and Mac OS.

JasperHG90:
Got it! Thanks for the explanation.

PR #46, opened 2017-08-10
One line fix for readSurvey

EruIluvatar:
Logics is hard, it seems...
Only change encoding to UFT-8-BOM if legacyFormat is set and **no** fileEncoding is specified.


JasperHG90:
Thanks!

Issue #35, opened 2017-06-26
Include question label in output of getSurveyQuestions()


vince-p:
It would be great to also have a column showing the question label for each item.
I.e., include:
QID
Question Label
Question Text

This is a really handy function - thanks!

JasperHG90:
Hi Vince,

Thanks for your request. I'll see if that's included in the result!

Best,

Jasper.

KAnalytics:
Hi Jasper,

It seems like the question label using getSurvey() is slightly different than using getSurveyQuestions(). I think this affects matrix based questionnaires that have many sub-questions.

Thanks,

Kai

JasperHG90:
@phoebewong fixed this.

PR #52, opened 2017-10-23
Add question label and force response output in getSurveyQuestions()

phoebewong:
I added question label in response to issue #35 in the output of `getSurveyQuestions()`. 

I also added force response (logical) in the output because I personally found it helpful as I sometimes use that to check if all questions have force response validated before publishing the survey. Let me know what do you think. Thanks!

### DIFF
--- a/R/getSurveyQuestions.R
+++ b/R/getSurveyQuestions.R
@@ -52,11 +52,13 @@ getSurveyQuestions <- function(surveyID) {
   resp <- qualtricsApiRequest("GET", root_url)
   # Get question information and map
   qi <- resp$result$questions
-  # Add questions
+  # Add questions, question labels and force response info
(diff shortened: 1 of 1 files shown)


phoebewong:
Working on fixing the error

JasperHG90:
Thanks!

phoebewong:
Hi @JasperHG90, I actually meant that I was working on fixing the error, since some tests still failed and the coverage decreased. The error was not fixed, but it is now fixed now. Should I make a new merge request now? Thanks!

JasperHG90:
Ah! I shouldn’t do these things late at night ... doh! Yes please do!

phoebewong:
No problem! I submitted a new PR, although it looks like it still failed the test.. Is it just because I haven't written a test for it?

PR #57, opened 2017-11-10
readSurvey: Use UTF-8 by default for current format

EruIluvatar:
Qualtrics provides the current csv format with UTF-8 encoding, so better set it by default.


JasperHG90:
Thanks for the suggestion!

PR #86, opened 2018-07-31
Don't treat NA strings as NA values. Fixes #85

### LINKED ISSUE #85: Answers of "NA" dropped by readSurvey()
Survey answers of "NA" are set to actual NA by readSurvey(). This is because readr::read_csv() defaults the na parameter to c("", "NA"). 

Specifying c("") for the na parameter resolves this issue.

And, yes - some of my respondents actually answer with "NA" instead of skipping a question :)

nickv9:
fixes #85 
Respondents occasionally supply answers of "NA" instead of skipping questions.  These answers are dropped by qualtRics because it calls readr::read_csv with the default value for the na parameter.

NOTE: setting the quoted_na to FALSE does not resolve this because the answer strings are not quoted.


JasperHG90:
Thanks!

PR #93, opened 2018-11-08
returns text instead of error when responses=0; fixes #92

### LINKED ISSUE #92: readSurvey() "Error: The survey you are trying to import has no responses."
I'm using getSurvey() in a loop to gather data from multiple surveys. If it gets to a survey with no responses, readSurvey() returns an error and terminates. Instead of using assert_that to check number of responses and error if 0, I would the function to return a warning, or nothing at all, or a data frame with 0 records. Then I can choose what contingency I want to perform.

Scott123456:
A possible solution for #92. Returning something always seems better than returning an error. Then the user can develop their own contingency.

### DIFF
--- a/R/readSurvey.R
+++ b/R/readSurvey.R
@@ -66,8 +66,9 @@ readSurvey <- function(file_name,
                              skip = skipNr,
                              na = c("")))
   # Need contingency when 0 rows
-  assertthat::assert_that(nrow(rawdata) > 0,
-                          msg="The survey you are trying to import has no responses.") # nolint
+  if(nrow(rawdata) == 0) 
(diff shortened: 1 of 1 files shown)


juliasilge:
Let me know if you are interested in revisiting this @Scott123456! For any future users running into this issue, I recommend using [purrr::safely](https://r4ds.had.co.nz/iteration.html#dealing-with-failure) for iteration.

Scott123456:
Hello @juliasilge. Weird I never received any notification you were responding to me. Anyways, I came back to see whatever happened with this, and lo and behold it was fixed just a few days ago. Thanks! And I agree that returning a tibble is much preferred. I will test it out.

Issue #89, opened 2018-09-24
Error in if (cnt$OK) { : argument is of length zero


jzadra:
I'm getting `Error in if (cnt$OK) { : argument is of length zero` when trying to use `getSurveys()` or `getSurvey()`.  There is no further explanation.  

I have tried using the yml config file as well as `registerOptions()` and have ensured that I"m using the proper token and base url.

ck2136:
I'm having the same problem.

Just for reference here's my session info

``` sessionInfo()
R version 3.5.1 (2018-07-02)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.1 LTS

Matrix products: default
BLAS: /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/libopenblasp-r0.2.20.so

locale:
 [1] LC_CTYPE=en_US.UTF-8      
 [2] LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8       
 [4] LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8   
 [6] LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8      
 [8] LC_NAME=C                 
 [9] LC_ADDRESS=C              
[10] LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8
[12] LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets 
[6] methods   base     

other attached packages:
[1] rstudioapi_0.8 boxr_0.3.4     qualtRics_3.0 
[4] nvimcom_0.9-58 colorout_1.2-0

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.19      pillar_1.3.0     
 [3] compiler_3.5.1    bindr_0.1.1      
 [5] forcats_0.3.0     tools_3.5.1      
 [7] jsonlite_1.5      tibble_1.4.2     
 [9] nlme_3.1-137      lattice_0.20-35  
[11] pkgconfig_2.0.2   rlang_0.3.0.1    
[13] curl_3.2          yaml_2.2.0       
[15] haven_1.1.2       bindrcpp_0.2.2   
[17] httr_1.3.1        dplyr_0.7.7      
[19] stringr_1.3.1     hms_0.4.2        
[...]

PR #103, opened 2019-03-20
Partially addressing #89, add error message for 403's

jmobrien:
You get "Error in if (cnt$OK) { : argument is of length zero" if the server returns an unspecified status_code.  403 is one, associated with API permissions on otherwise valid tokens.  

Fixing this more generally would probably involve a few lines that notes if the status code doesn't fit any specified type, and returning whatever that code is for diagnostics.  Otherwise it's ambiguous.


juliasilge:
Thanks so much for the PR @jmobrien! 🙌

PR #108, opened 2019-03-27
Add GitHub link to DESCRIPTION

maelle:
I think it might be best to keep both, for metadata gatherers (e.g. pkgdown only adds "browse source code at" on the right if the GitHub repo is there, even if there's an issue tracker).

Btw I added the URL of the pkgdown website near the repo description :wink:


juliasilge:
Thank you! 🙌

Issue #130, opened 2019-10-20
Move to new version of API


juliasilge:
Users are starting to have various problems, such as those noted in #94 and #121, that I believe are related to being on the [v3 `responseexports` API](https://api.qualtrics.com/docs/response-exports).

Looks like it's time to bite the bullet and switch to the [new export API](https://api.qualtrics.com/docs/getting-survey-responses-via-the-new-export-apis). This will impact most functionality in this package. When we make this switch, I plan for this package to no longer support the older "Legacy" format.

chrisumphlett:
+1 here. I too would like to utilize the progress and duration columns mentioned in #94

jmobrien:
Not sure if this is useful, but I have a working version of fetch_survey for the new api in a fork.  Needed the new question randomization variables for a project.  

Still needs some attention, but it works as a proof-of-concept and with what feels like a fairly minimal change from the existing codebase.  

It does include the progress and duration columns mentioned in #94, but I can't speak to whether it addresses the issues raised in #121.

jmobrien:
I should also note that in the rework I ended up changing `append_root_url()` to `create_root_url()` and making a few other separate utility functions that further build on it (like `create_fetch_url()`.  I did hate to make such general changes, but the previous approach didn't quite fit when all survey-related API queries have urls that include `/surveys/` now.   `root_url` just terminates in `/API/v3/` now.  

I only bring this up as the last time I worked with code from this package was to repurpose it for fetching mailing lists and pushing back minor edits to them. Similar to #131 and @dsen6644's (arguably cleaner) example*.   As @dsen6644 noted and I experienced as well, you have to manually work around the current `append_root_url` to add any additional capabilities.   Conceptually grounding on a `root_url` that's a top-level address only might be a small change that sets up building out this sort of additional functionality in the package.  

*notably, both of us used the [API for mailing lists](https://api.qualtrics.com/reference#list-mailing-lists) in standard libraries and not [XM directory lists](https://api.qualtrics.com/reference#list-mailing-lists-1) that was the main topic of #131.  But extensions to either list would work similarly.

juliasilge:
Well, this made my day @jmobrien! 🤩 Thank you so much for working on this and for sharing what you've learned so far.

I believe you are making the right decision about the changes to how the URLs are built, given the new endpoints. I will take a look at it but sounds right. 👍 

[...]

Issue #120, opened 2019-07-22
Update/improve testing


juliasilge:
Look into using [vcr](https://github.com/ropensci/vcr) for testing API calls.

juliasilge:
rOpenSci has a [resource here on using vcr, webmockr, etc for testing](https://books.ropensci.org/http-testing/).

shaun-jacks:
I could look into this, currently liking the idea of using webmockr and adding tests for behavior on different stubbed response codes

juliasilge:
I would definitely love help in moving the existing tests over to this more modern framework, and making sure that the main user-facing functions are adequately tested. 🙏

juliasilge:
Closed in #140

PR #135, opened 2020-02-12
Changes to support using the newest API for fetching survey data, addressing #130

jmobrien:
- fetch_survey now points to the modern API endpoint
- swapped/added some of the new parameters, not everything yet: 
     https://api.qualtrics.com/reference#create-response-export-new
- updated some assertions & associated utilities to support argument checking, etc.
- url building redesigned to work better with new API, may be slightly more easily extensible going forward

Generally, `fetch_survey` seems to work, and a few internal changes were made so `metadata` wasn't broken by this.  But remains to be seen if there's deeper stuff that needs addressing/cleaning up.


jmobrien:
Updated 2019/02/13 to add a fix to all_surveys() that needed to work with the new URL generation scheme. (633ca33)

juliasilge:
A question for you @jmobrien: did the option in the API for `unanswer_recode` really change to something integer-like? Not a string anymore? I am looking in the documentation trying to find this but not coming upon it easily.

juliasilge:
Wait, [I think I found it](https://api.qualtrics.com/reference#create-response-export-new).

juliasilge:
OK! I think this does it. 🎉  Thank you so much for your work on this.

I had to remove one test because the old testing framework does not play nicely with the new API calls. This makes #120 a higher priority.

chrisumphlett:
Thanks to @jmobrien -- I'll give it some hardcore testing very soon, perhaps even tomorrow if I have time.

jmobrien:
Yes, it's specified as int32 now.  Can't recall if I still tried it as a
string, but I did see the API fail with a non-integer number, so I assume
it's not at all flexible.

Joseph O'Brien, Ph.D.
Adolescent Development Research Group
Department of Psychology
University of Texas at Austin
joseph.obrien@utexas.edu

jmobrien:
Glad to help.  I've used it for ~400 data downloads so far, albeit with
consistent parameters other than survey IDs.  So far everything looks good,
but I'll check in if anything looks off.

Joseph O'Brien, Ph.D.
Adolescent Development Research Group
Department of Psychology
University of Texas at Austin
joseph.obrien@utexas.edu

chrisumphlett:
I think something changed... I"m not quite sure what. This is my process:
* `fetch_survey`, put in df called `responses`
* extract the column names and labels with: `col_name_labels <- stack(sjlabelled::get_label(responses)) `
* this creates a df with columns `values` and `ind`.

Previously, I then had logic that filtered on `values` expecting to see the column name there. Now, I'm having to switch that filter to reference the `ind` column.

Am I making sense, and does that change in behavior make sense according to the change? Not saying it's a problem!

Context:
The goal of this process is to dynamically build a response_metadata table where I log all of the response metadata (start date, finished, **progress** now that it's there (yay!), browser) into a long table, for all of the surveys.

juliasilge:
Thanks for checking things out @chrisumphlett! 

For the browser info, it looks like that is an actual difference between the old endpoint and the new endpoint, i.e. what Qualtrics is sending over when we request data.

For the labels, I think there may be some digging still to do, to figure out what the best end state for label handling is. If you would like, can you chime in on #134 about issues with the labels/attributes?

chrisumphlett:
Re: Browsers. Makes sense to me-- I have no use for that information anyway, just need to update code.

I will comment on #134 momentarily.

PR #137, opened 2020-02-14
Fix for issue with time_zone

jmobrien:
In the new `fetch_survey`, time_zone works as expected in terms of the raw data returned from Qualtrics, but `tidyr::convert` in `read_survey` was then processing dates/times to the user's `default_local()` time, regardless, leading to errors in time variables where they were discrepant.

For example: my `default_local` was UTC, but I fetched with `time_zone = "America/Chicago"` so my times in the downloaded data would be marked consistent with where collection happened.  Data displayed the correct local times, but **in UTC**, which was therefore 6 hours off on the actual POSIX data.  

(Possibly happening since `read_csv` replaced `read.csv` last year?)

Quick fix was to pass the `time _zone` parameter to `read_survey` and subsequently `readr::convert`.  Appears to work; not sure if there's a better approach.

(Also, 99dbd8c addresses bug in assertion check for unanswer_recode_multi, which it looks like @juliasilge  already caught, thanks.)


juliasilge:
Thanks so much @jmobrien! 🕰✨⏰

PR #141, opened 2020-02-20
feature(breakout_sets) added breakout_sets parameter that also exists…

shaun-jacks:
… in API/v3 docs

This allows all multiple choice column answers to either be split to columns or collapsed to one column per multiple choice question. This saves a lot of manual data processing since it's exposed in the API. I defaulted the parameter to TRUE to stay backwards compatible with original responses. It could probably also default to NULL.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@
 - Deprecate old versions of functions: `getSurveys()`, `getSurveyQuestions()`, `getSurvey()`, `readSurvey()`
 - Move to updated version of Qualtrics API (#130) thanks to @jmobrien
 - Correctly handle time zone conversions (#137) thanks to @jmobrien
+- Add `breakout_sets` parameter thanks to @shaun-jacks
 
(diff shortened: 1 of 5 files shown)


11rchitwood:
@shaun-jacks can you link to this feature in the [API reference](https://api.qualtrics.com/reference)?

shaun-jacks:
Yes I can! Here's the link

https://api.qualtrics.com/reference#create-response-export-new

If you scroll down you'll see the breakoutSets parameter. Toggling it to true or false will control the multivalue fields export

11rchitwood:
Thanks @shaun-jacks ! I definitely think this would be useful feature.

juliasilge:
Thanks so much for this @shaun-jacks!! 🙌 If you are interested in implementing new testing features, described in #120, I am looking for some help on that!

shaun-jacks:
Hi @juliasilge  and @11rchitwood, thank you and of course! I would love to help on my spare time :)

PR #146, opened 2020-03-18
new column mapping functionality

jmobrien:
1. Obtaining column mapping data from w/in `fetch_survey` (mostly `read_survey`) , added as attributes to response data columns.  Might be extensible to do something like `column_map` under the new system.

2. New (unexported) function `extract_colmap()` may for now be useful for making column maps from existing downloaded responses.

3. Few related and unrelated additions to argument checking.  

4. Incidentally ended up fixing the use of QIDs in `read_survey` via param `import_ids`, which had apparently broken with migration to new API w/`fetch_survey`.  

5. Added new checks for column mapping with legacy formats, and updated import_id checks as that was reworked a bit.  Built both as warnings rather than errors; ignore/revert if not preferred.  "Legacy" is actually 2 versions back at this point, not sure how relevant it remains.


jmobrien:
Added an additional commit fixing some import_id's issues.

juliasilge:
Thank you so much for this PR @jmobrien! 🙌 I pushed a few changes so that this will pass R CMD check.

The main thing here that we need before merging are some tests. I know that the existing testing suite is in rough shape, and there is no need to redo the whole thing at one go. However, to add new functionality like this, we need to add a test or two along with it. We don't want to keep using the outdated testing frameworks for API calls that are currently there; instead, we want to move over to vcr and webmockr. rOpenSci [has a great resource here](https://books.ropensci.org/http-testing/) on how to get started with tests for API calls.

Can you add a test or two to this PR so I can merge it?

jmobrien:
@juliasilge I'd be glad to add some appropriate testing, and already tried to learn a bit after noticing a similar comment associated with the great new mailinglist features from @dsen6644.  Also, the page you linked looks great when viewed as technical implementation information, so thatnks. 

Unfortunately, the linked pages don't really bridge the gap that I have, which is that I'm a psychologist, not a CS-type, so I'm not really in touch with the broader philosophy of (unit?) testing and its goals. So, I'm not really clear what I'm aiming for here.  I still want to help, but bear with me a minute as I have some questions.  

1. I do generally appreciate the idea of creating reprexes (see, e.g., [here](https://github.com/glmmTMB/glmmTMB/issues/568#issue-589445564), reworked as a test by the maintainer [here](https://github.com/glmmTMB/glmmTMB/pull/569/files)).  But my understanding of their use is more contained to illuminating a specific *problem* or bug, then later perhaps re-using that code to demonstrate that any fixes are working correctly.  But, compared against a reprex or a specific bug, a new *feature* like a column map obviously has many aspects that together need to work to make the output "correct."  So, I'm not sure what makes a test adequate here.  

2. More specific to this PR, both your comment above and most of the content that you linked on `webmockr` and `vcr` seem to focus on testing API calls specifically.  That's even more outside my area--literally 100% of what I know about APIs comes from me trying to reverse-engineer things from this package--so I'm even less in touch with best practices, and what a "good" API test should try to accomplish.  It's also a bit confusing for me in context of this specific PR, as I don't think it touches anything API-related--IIRC, I just reworked how we were processing some response content that had already been saved to disk.  So, I would love to know what we're aiming for here that's needed prior to merging.  

(I did of course also make the new `fetch_survey` that retrieved said content, which DOES touch our API functionality, so I'd also like to know if we need tests for that, and if so, what those should look like.)

Apologies for the length, I'm really just not sure how best to be helpful here, and I'm trying to give enough detail that I can work out with you how to bridge the gap.

jmobrien:
@juliasilge Leaving aside the above, a comment on this PR itself: having used its functionality for my own work a bit now, I'm wondering if it could use some additional adjustment.  

It works "correctly," in the sense that it's grabbing all the various elements provided in the qualtrics-provided JSON row as attributes/columns. But the data in two key attributes/columns, `ImportId` & `ChoiceId`, are a bit more difficult to interpret than you might think.  Depending on context, sometimes the `ChoiceId` is actually a "choice" ID number (as defined internally by Qualtrics, and therefore in the "get survey" content),  but sometimes it's instead an "answer" ID number, user-specified question name, or flow ID.  And sometimes the "choice" isn't in that column, but instead appended as a suffix to the ImportId column.  

This is all qualtrics's doing in how they formatted their response data.  And I think it IS possible to work everything out once you know what you're looking at/for.  But as a "column map" it's a bit confusing for end users.  Not sure how much the core goal of the package is to provide access to explicitly what data Qualtrics makes available, versus providing access to more human-readable formats, but that might speak to goals here.  

BTW, related to this in #151: does column `Q1_DO_1` in your example relate to a matrix question block that used randomized row display?  We generally only randomize at higher levels, so I hadn't encountered a name like that (i.e, [QID]_DO_[number]).

juliasilge:
Thank you so much for all the work you have invested already @jmobrien! 🙌 

I too do not come from a classic CS background so I totally get where you are coming from. I found [this section of the R packages book](http://r-pkgs.had.co.nz/tests.html#test-tests) helpful as a starting place for thinking about what to test.

In terms of what test(s) to add for this PR, I think what we need is a test to confirm that `fetch_survey()` correctly handles `colmap_attrs = TRUE` and `colmap_attrs = FALSE` and returns the correct result for an example survey (stored in a mocked API call). You could perhaps write a test for `extract_colmap()`, testing that it returns the result you expect for some example data. You could check out the [tests in ropenaq](https://github.com/ropensci/ropenaq/tree/master/tests/testthat), which are set up generally the way I would like. And the existing tests [aren't **all** bad](https://github.com/ropensci/qualtRics/tree/master/tests/testthat).

The issues with Qualtrics and choice vs. IDs has been quite a challenge for users in general (and not just the R users). I tend to lean toward making a few careful choices in the package about "adjusting" info but generally giving users that Qualtrics API results. Let me know if you have more thoughts after using this for a while.

(And yes, I believe in #151 those were a matrix of questions that were randomized.)

jmobrien:
Thanks.  Just to make sure I'm tracking, I'm reading that we still wanting to use the `devtools` testing framework, just with something saved via webmockr/vcr so tests can run w/o actual API calls, correct?  If so, I'll look back over webmockr/vcr above with that in mind, so I can see best how to do that.  I'm familiar with Hadley's R packages book, but I hadn't looked closely at testing; thanks for that resource!

Some design questions around column mapping and specifically `colmap_extract()`, though: what are your thoughts on this workflow design?  There's a reason I didn't export it and left it just as a proof-of-concept that we could get similar results to `column_map()`.  I think there are a few things to think through:

1. `labels = TRUE/FALSE` : If we're taking #134 seriously, this approach adds several more variable attributes people might not want, so they'd have to set `colmap_attrs = FALSE`. But then they couldn't use `column_extract()` subsequently.    So, it if you DO want a column map, but DON'T want a bunch of attributes that print, it's not ideal.  And attributes can be annoying to remove.

2. Partly because the labels (i.e., item text) are there, right now `column_extract` is liberal and grabs all non-essential attributes to put in the column map. Would be easy enough to make it more explicit, but should we include labels? I'm inclined to say yes b/c then it also works like a more directly informative `survey_questions()`, but then the questions about parameters for including excluding labels in #134 become relevant for different reasons.

Partly for all this, I'm wondering if we should instead, or as an option, go ahead and construct the column map up front, then add it as an attribute to the whole data frame.  It's cleaner, and allows for having column maps w/o extra attributes.  Extraction would then be a breeze.

We could maybe still include the option for variable-level attributes (could be nice to be able to, say, construct a column map after subsetting, if you wanted), and let the user decide what they want.  But I'd love your thoughts.

jmobrien:
Also noting here: current column mapping produces incorrect results when `fetch_survey` pulls a 0-response survey, related to how `read_survey` terminates early under those conditions.  Will be fixed and an appropriate test created.

jmobrien:
Additional note, adding `convert = TRUE` on `fetch_survey()` drops existing attributes so everything gets lost.  Didn't notice as I don't use that feature myself, but think this maybe suggests the need for a bigger review/rework. 

How should I handle this PR?  Close it and resubmit after the changes + tests are done?  Just leave this up and add commit to the colmap branch as I go?  (Apologies, I'm still learning the default protocols/culture surrounding git collaboration.)

Also, @juliasilge I expect this will have me looking over both parts of what's being discussed in #134, i.e., the question text as "label" attribute and the qualtrics "labels" that is what `convert = TRUE` uses.  So, if there's a decision in the near future surrounding the UI for those two aspects, wouldn't likely be much trouble to handle that as I go here.  It will be a few days at least before I can return to this in earnest, so just let me know.

juliasilge:
Thanks for your patience @jmobrien, and all your excellent work! 🙌 

I think it would be fine to leave this PR open and continue to push commits to it, unless that becomes inconvenient for you, in which case it is fine to close and open a new one.

- Yes, we definitely will continue to use devtools, but with webmockr + vcr to store "mocked" API calls to use as data in the tests. [Some of the tests](https://github.com/ropensci/qualtRics/blob/5da0bd16d56642d26ab6fd2f87f29f2e6d56d14d/tests/testthat/test-all-surveys.R#L4) currently use an older, less awesome way to do this via the httptest package. We don't need to get rid of the old stuff now but any new tests should should use webmockr + vcr as outlined [in this resource](https://books.ropensci.org/http-testing/).

- I have been looking at both this and the issue with #134 and I wonder if the answer is to add this info as an attribute to the whole data frame (rather than the columns), like you mentioned. I like this idea a lot. We'll have to look at what `convert = TRUE` is doing if so.

jmobrien:
Currently, the column map functionality all happens within `read_survey()`, while `convert = TRUE` does occurs in `infer_data_types()` called just after in `fetch_survey()` (see [here](https://github.com/ropensci/qualtRics/blob/5da0bd16d56642d26ab6fd2f87f29f2e6d56d14d/R/fetch_survey.R#L179-L184)).  So, currently, I think the conversion is just just replacing variables with new ones, which doesn't preserve any previously-applied attributes.  

Unless I'm mistaken, that means that if we create a column map as a dataframe attribute applied to the response dataframe, then the conversion issue would resolve itself?

Do we want to include labels in the column map?  I like that organization, but I think the "labels" attribute relates to being able to export data to Stata, etc., so whether we (also) keep it as an attribute should likely be considered on its own merits.

jmobrien:
Related: column map creation was initially implemented in `read_survey` because it leverages code used when `import_id = TRUE` that loads up the JSON row from the temp file to rename the data ([here](https://github.com/ropensci/qualtRics/blob/5da0bd16d56642d26ab6fd2f87f29f2e6d56d14d/R/read_survey.R#L102-L113)).  But we could just as easily make that its own utility function that reads the same temp file, and call it from `fetch_survey` That might have better modularity, and minimize issues of competing functionality over further development.

jmobrien:
@juliasilge Just adding an update to say that this hasn't been forgotten, I'm just not had the time to come back to this while I focus on my main project.  Looks like other folks have done some great work getting webmockr, etc. integrated (h/t @dsen6644 & @shaun-jacks !), so maybe I can trade on that to quickly complete this later.

Just FYI, I've also had for a while a functioning update of `metadata()` that queries the new v3 metadata endpoint, which is a better match to the current v3 `fetch_survey()`.  Been holding off on pushing that b/c that obviously will need tests, but I should be able to do both together and then offer them up.

juliasilge:
An updated `metadata()` would be fantastic. 🙌  If it's easier to submit that from a new PR (branched clean from master) feel free to do that. Whatever works best for you!

jmobrien:
Yeah, they were previously separate branches on my fork, but then I ended up combining them so I could install and use a package version that had both.  But I agree separate PRs is best, so when it comes time I'll do it that way.  Just saying I might be able to complete them on a similar timeline.

As far as v3 metadata, right now I have `metadata2()` that queries the new endpoint and takes the complete list-formatted JSON, much like the base functionality of its predecessor when everything is added to `get`. The content and structure is a bit different from the old version, though, so we can't 100% transfer the old functionality.  Before I proceed I'll raise a new issue so folks can discuss what they actually want the new function to do.

jmobrien:
Finally updating this with tests and slightly improved functionality.  Still getting familiar with vcr/webmockr, so my records look a bit different than others (e.g., it uses "<<my base url>>" instead of "t.qualtrics.com" as seen in other .yml files).  I'm not quite sure I'm implementing the workflow correctly, but I think what I have is roughly correct.  The tests passed locally.

Just a note: I'm getting "Namespace dependency not required: ‘tidyr’" from the local checks, which I'm not sure how to interpret.  Those were added 

Quick recap since it's been a while:
1. Adds functionality to `fetch_survey()`/`read_survey()` that adds a column map to the response download, attached as an attribute.
2. The function `make_colmap()` does the heavy lifting here, pulling this all from the JSON row of the response download.  Right now this is in utils.R, not sure if that's where it belongs.
3. Before attaching to the dataframe, a variable `description` also gets added to the column map.  It's the same thing as what is added to variables as a "label" attribute right now, but without separating by subquestion (IMO we should change the labels to work this way too, see #168) 
4. Creates `extract_colmap()` as a simple helper function that pulls out the attribute as an independent data frame.  In theory this could be folded into the existing column_map, or expanded to do a v3 API call on its own.  
5. Also implements `[QID]_[choiceId]` as the variable names when using `import_id = TRUE` to ensure uniqueness, generating those from the generated column map.  Since I already had it this way, it ends up replacing @lyh970817's update, which does the same thing in a rather similar, but self-contained way.  His solution could be kept instead if preferable.

jmobrien:
Closing in favor of #199, which is better organized and more comprehensive.

Issue #153, opened 2020-04-12
last_response still fetches all responses


weiyangtham:
I'm otherwise able to get the responses with `fetch_survey`, but even when I enter a response ID into  `last_response`, I get all the responses instead of only those after the response ID. I've tried those on a couple of different projects without success

juliasilge:
Well. This is rather embarrassing. 🙈 

I haven't noticed this since taking over maintenance over this package, but the `last_response` argument doesn't actually do anything. The argument [is checked](https://github.com/ropensci/qualtRics/blob/6f5c0c5e39657464eb2402493a8d730643b0c805/R/fetch_survey.R#L108) but then never actually used. Specifically, it is [not included in the JSON payload](https://github.com/ropensci/qualtRics/blob/6f5c0c5e39657464eb2402493a8d730643b0c805/R/fetch_survey.R#L146) sent to the Qualtrics API. This is probably because this is [not one of the parameters of the API](https://api.qualtrics.com/reference#create-response-export-new).

So where does that leave us? I am going to need to deprecate this argument because the API doesn't work like this. Thanks for bringing this to my attention, and I apologize for the confusion!

weiyangtham:
Good to know and thank you! the package is lifesaver

juliasilge:
Just found this in `check_param()`:

```r
  # No longer used:
  # if ("last_response" %in% names(args)) {
  #   if (!is.null(args$last_response)) {
  #     assert_lastResponseId_string(args$last_response)
  #   }
```

juliasilge:
The warning users will see will look like this:

```out
Warning messages:
1: The `last_response` argument of `fetch_survey()` is deprecated as of qualtRics 3.1.2.
This warning is displayed once every 8 hours.
Call `lifecycle::last_warnings()` to see where this warning was generated. 
```

jmobrien:
Sorry, @juliasilge.  You should brush off that embarrassment and send it my way.  

The older response API endpoint DID have had a `lastResponseId` param that this got passed to, and I neglected to document the change when it was removed in #135.  

The new API doesn't support anything quite like it, though.  There are "continuation tokens" but we don't support them yet, and they require some forethought to work.  

However, @weiyangtham, I *think* you could get the same effect by pulling RecordedDate from the last response and passing that to `start_date`.  

You'd need to format it properly, though, see here:
https://api.qualtrics.com/docs/dates-and-times

PR #155, opened 2020-04-17
using jsonlite to automatically build payloads

jmobrien:
Suggested change--still needs tests written, but thought I'd go ahead and put it in the pipe for review.  So far seems to work fine.

- Uses `toJSON` from package `jsonlite` for automated construction of payloads, eliminating the massive manual build using a bunch of `ifelse`s and `paste`.  `qualtRics` has imported functions from `jsonlite` for quite a while now, so this doesn't add any new dependencies.  

- `purrr::discard` drops any null parameters when building, as qualtrics API doesn't like actual json `nulls` or empty sets.

- Retains original technique for formatting`start_date` and `end_date` to fit ISO 8601 as expected (https://github.com/ropensci/qualtRics/compar/master...jmobrien:payload?expand=1).  Looking ahead, though, could be easier now to set up more flexible handling of start and end times, useful now that since-last-response request functionality is gone, see #153.


juliasilge:
This is fantastic @jmobrien! Thank you so much. 💯

jmobrien:
Glad it helped.  Looking forward, we might want to think about how best to have this tool accept the standard date/time output used by the `readr` tools we're currently employing in `read_survey`.  I played around with it for a bit earlier before submitting but didn't see a standard tool that could do date+time with the `"T"` separator, plus I wasn't sure how to implement timezones in a way that made the most sense and properly considered how the `useLocal` flag was set.

juliasilge:
If you want to open an issue with your ideas and any summary of what you tried, we can save that for later.

jmobrien:
Can't say I spent a ton of time on it, but I'll write a few lines.

PR #162, opened 2020-05-10
Col_types argument

jntrcs:
Here is my stab at allowing people to have more control over how the data is read in. I added col_types as an argument to `read_survey()` and `fetch_survey()` (which passes it to `read_survey`).

I'm more confident in the `read_survey()` addition than `fetch_survey()`. `fetch_survey()` does conversion after it calls `read_survey()`, so it gets a little hairy about what the conversion rules hierarchy should be. I added the details in separate commits, so you can hopefully sort through them and accept only the `read_survey()` one if you wish. 

I added a couple assertions to the test cases to ensure we're getting the data types we want.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -4,6 +4,8 @@ Title: Download 'Qualtrics' Survey Data
 Version: 3.1.2
 Authors@R: c(person("Jasper", "Ginn", role = c("aut"),
              email = "jasperginn@gmail.com"),
+             person("Jackson", "Curtis", role = c("ctb"),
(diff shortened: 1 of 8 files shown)


juliasilge:
This looks fantastic @jntrcs! 🙌  Would you like to add yourself as `"ctb"` to DESCRIPTION before I merge this in? We are in alphabetical order by last name.

jntrcs:
Added! Thanks a lot. Contributing to CRAN has been a bucket list item for awhile. Now that I've got my feet wet hope to dive in more often.

PR #169, opened 2020-05-19
`fetch_distributions` to pull distribution data for a specific survey

dsen6644:
Returns all elements provided by the api. Definitions of each field can be found [here](https://api.qualtrics.com/docs/getting-information-about-distributions-1#get-distribution).

Right now, the object returned is a single large data frame (35 fields). However, I am open to keeping the raw api structure and returning a list of data frames instead (ie. x$stats_ blocked becomes x$stats$blocked).


juliasilge:
Thank you so much @dsen6644! 🙌  Is it possible for you to check in a test fixture `fetch_distributions.yml` to go along with your test?

dsen6644:
Certainly! Few questions...
- Is `qualtrics_api_credentials(api_key = "1234", base_url = "t.qualtrics.com")` an actual account? It fails when I try to connect. Or are you using a separate account to record the vcr and the above credentials are dummy variables?
- All of the vcr tests you've generated have the body encoded as `ASCII-8BIT` were the ones I produce are `UTF-8`. Did you specify this somewhere?

juliasilge:
I described how I created the test fixtures [in this comment](https://github.com/ropensci/vcr/issues/167#issuecomment-625553365). Take a look and see if that makes sense. Basically, I created them **interactively** in the console using the vcr functions, not using `devtools::test()`.

I believe the reason the encoding is different is because I used this option in `vcr_configure()`:

https://github.com/ropensci/qualtRics/blob/fecbb88331de340f050138dd666ed226cfce4d4c/tests/testthat/helper-qualtRics.R#L5

I ran that function interactively to set things up and get the fixtures saved in the right directory, but unfortunately I could not get the fixtures created automatically via `devtools::test()`.

dsen6644:
Thank you! And as for the first question? Is `base_url = "t.qualtrics.com"` a real rest account? Or strictly a dummy string?

Also, I have `preserve_exact_body_bytes = TRUE` in my `vcr_configure` as well, still getting an encoding of `UTF-8`. I'll keep looking into it, not sure why the same configuration is producing different outputs.

juliasilge:
Ah, `base_url = "t.qualtrics.com"` is just a dummy string, like the API key.

When I created the test fixtures interactively, I used my real API credentials. These test fixtures then had the sensitive info stripped out and replace with `<<<my_base_url>>>` and similar. The line

```r
qualtrics_api_credentials(api_key = "1234", base_url = "t.qualtrics.com")
```
is used in the test to store credentials in the environment while a test is run. The [function here](https://github.com/ropensci/qualtRics/blob/master/tests/testthat/helper-qualtRics.R) replaces those values like `<<<my_api_key>>>` with the environment variables while a vcr test is run.

dsen6644:
I was able to get the `test-fetch-distributions` to pass a local `devtools::check` only with MY api credentials set via `readRenviron("~/.Renviron")`. Using `qualtrics_api_credentials(api_key = "1234", base_url = "t.qualtrics.com")` results in a failed test. Also - getting failed tests for many `fetch_survey` tests. The `vcr` workflow has been challenging to consistently replicate.

juliasilge:
You are soooooo correct about vcr and its challenges. We are somewhat early adopters and I know they are still working some kinks out; thank you for your 💯 patience in hanging with me through getting this done. 

The testing situation was sort of dire before so believe it or not, this is actually an improvement, albeit frustrating. 🙈

PR #170, opened 2020-05-20
Update read_survey.R

seanchrismurphy:
Fix parsing of ImportIds for column names. The previous gsub wasn't capturing properly so the names were being read as JSON. But even when fixed, sometimes there are more than one element in the JSON array for each variable. This approach reliably captures all importIDs, but does rely on JSONLite.

### DIFF
--- a/R/read_survey.R
+++ b/R/read_survey.R
@@ -117,7 +117,9 @@ read_survey <- function(file_name,
       n_max = 1
     ))
 
-    names(rawdata) <- jsonlite::fromJSON(paste0('[', paste(as.character(unlist(new_ids)), collapse = ','), ']'))$ImportId
+    names(rawdata) <- jsonlite::fromJSON(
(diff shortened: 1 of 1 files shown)


juliasilge:
This looks great! 🙌

juliasilge:
Did this solve the issue you had opened? Or was that unrelated? (I could not reproduce that problem with my own surveys.)

seanchrismurphy:
It didn't directly solve it, but it enabled me to create a workaround. Turns out the ImportIds -are- capturing the question recoding where the header is not. That is, take a 3-item matrix question (say Q13/QID20), with the items recoded as 3, 7, and 9. The header will be Q13_1, Q13_2, Q13_3. But the import ids will be QID20_3, QID20_7, QID20_9. 

If you can't reproduce that behavior on your account, I could try exporting a .qsf file from a survey that produces this behavior.

PR #175, opened 2020-06-07
`flatten_df` to flatten embedded data object before binding

dsen6644:
Embedded data fields to be individual columns rather than single list-column. This is more consistent with the mailing list export using the Qualtrics UI.


juliasilge:
Looks great! 🚀

Issue #181, opened 2020-06-29
fetch_survey() with import_id = TRUE does not have unique column names


lyh970817:
The behaviour seems a bit strange. There are some warnings from readr that say the column names are deduplicated but then in the survey data frame the deduplication didn't seem to take effect.

juliasilge:
Closed in #182

PR #182, opened 2020-07-02
Add choiceId to `read_survey` colnames. Addresses #181.

lyh970817:
Currently the colnames only use ImportId which will cause column name duplication.


juliasilge:
Thank you so much! 🙌

PR #191, opened 2020-10-26
Add parameter `survey_definitition` to request additional survey metadata

lyh970817:
This survey-defitition request is documented at https://api.qualtrics.com/api-reference/reference/surveyDefinitions.json/paths/~1survey-definitions~1%7BsurveyId%7D/get.

It seems to contain more information about questions and blocks (e.g. regarding the "loop and merge" option) but also lacks a few things that are in the original "surveys" request. I added a parameter `survey_definition`` but have not added information to the function documentation or provided a default value.

### DIFF
--- a/R/metadata.R
+++ b/R/metadata.R
@@ -52,6 +52,7 @@
 #' }
 #'
 metadata <- function(surveyID,
+                     survey_definition,
                      get = list(),
                      ...) {
 
@@ -103,7 +104,7 @@ metadata <- function(surveyID,
   # QUERY API ----
 
   # Function-specific API stuff
-  survey_url <- create_survey_url(Sys.getenv("QUALTRICS_BASE_URL"), surveyID)
+  survey_url <- create_survey_url(Sys.getenv("QUALTRICS_BASE_URL"), surveyID, survey_definition)
 
   # Send GET request to specific survey
   resp <- qualtrics_api_request("GET", survey_url)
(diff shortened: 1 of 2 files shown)


R/metadata.R
```diff
@@ -52,6 +52,7 @@
 #' }
 #'
 metadata <- function(surveyID,
+                     survey_definition,
```
juliasilge, line 55: Let's make this an argument called "endpoint" with two options for now that match the names of the API endpoints we're using. Let's use [`rlang::arg_match()` to match them](https://rlang.r-lib.org/reference/arg_match.html), and the endpoint we're using right now ("surveys") as the default, i.e. first.

R/metadata.R
```diff
@@ -103,7 +104,7 @@ metadata <- function(surveyID,
   # QUERY API ----
 
   # Function-specific API stuff
-  survey_url <- create_survey_url(Sys.getenv("QUALTRICS_BASE_URL"), surveyID)
+  survey_url <- create_survey_url(Sys.getenv("QUALTRICS_BASE_URL"), surveyID, survey_definition)
```
juliasilge, line 107: Use the new `endpoint` variable to create the URL; the `create_survey_url()` function can be changed to paste together the endpoint with the other info.

R/metadata.R
```diff
@@ -114,57 +115,91 @@ metadata <- function(surveyID,
...
-      NA,
-      resp_filt$expiration$endDate
+  if (survey_definition) {
+    metadata <- data.frame(
+      "surveyID" = resp_filt$SurveyID,
+      "name" = resp_filt$SurveyName,
+      "ownerId" = resp_filt$OwnerID,
+      "organizationId" = resp_filt$CreatorID,
+      "isActive" = resp_filt$SurveyStatus
+    )
```
juliasilge, line 125: There is a fair amount of repetition here and in this section. Pull out anything that is the same for both endpoints, out of the if statement, to make it easier to maintain this code.

R/metadata.R
```diff
@@ -114,57 +115,91 @@ metadata <- function(surveyID,
...
+    } else {
+      questions <- resp_filt$Questions
+    }
+
+    # Construct metadata
+    met <- list(
+      "metadata" = metadata,
+      "questions" = questions,
+      "blocks" = resp_filt$Blocks,
+      "flow" = resp_filt$SurveyFlow
```
juliasilge, line 146: This contains some variables (like `flow` and `blocks`) that are labeled `FALSE` in the `standard_list` in this function and don't get kept in what gets returned from this function. Is this intentional? It seems like you would want these, right? We'll need to carefully think through which objects to return.

juliasilge (changes requested):
Thank you so much for your great work here! 🚀 

This change to the `metadata()` function will definitely need a test or two. The tests for this function live in `test-metadata.R` and the fixtures live in `tests/fixtures/`. Like I mentioned on the other PR, I recommend looking through [rOpenSci's (draft) book on HTTP testing](https://books.ropensci.org/http-testing/) to get acclimated, and then giving a go at using [vcr](https://github.com/ropensci/vcr) to set up a test or two.

R/metadata.R
juliasilge: This argument will also [need to be documented](https://r-pkgs.org/man.html).

R/metadata.R
lyh970817: The lists for the two endpoints have slightly different names, for example the list for the "surveys" endpoint stores survey ID in `surveyID` whereas for the "surveys_definition" it is stored in `SurveyID`. So this is technically not a repetition?

juliasilge: Ah, I see what you're saying. When you get around to editing this PR, see if you can do any refactoring at all to improve this section, but if not we'll see what we can later.

juliasilge:
I believe we have handled the `survey-definitions` endpoint with the `fetch_description()` function. Thank you so much for this contribution! 🙌

PR #192, opened 2020-10-26
Make `import_id = TRUE` take effect when no response in survey

lyh970817:
Currently when there is no response in the survey, `import_id = TRUE` in `fetch_survey` does not take effect.

### DIFF
--- a/R/read_survey.R
+++ b/R/read_survey.R
@@ -88,17 +88,6 @@ read_survey <- function(file_name,
     n_max = 1
   )))
 
-  # Message for no data in survey
-  if (nrow(rawdata) < 1) {
-    message("The survey you are importing has no responses.")
-    tbl <- tibble::as_tibble(matrix(nrow = 0,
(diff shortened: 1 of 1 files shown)


juliasilge:
Thanks so much for this contribution! 🙌 

Can you write a test to go along with this change, to make sure that the correct column names get used for an empty survey? Testing in this package is kind of HARD MODE, because of the API requests, but want to give it a try?

The tests for this change should go in the `test-fetch-survey.R` file, and it will involve creating a test fixture. This package uses [vcr](https://github.com/ropensci/vcr) to create test fixtures that go in `tests/fixtures/`. I highly recommend checking out rOpenSci's [HTTP testing book](https://books.ropensci.org/http-testing/), currently in draft. Give a go at this and see if you can get a test working.

juliasilge:
Thank you so much for this contribution! The work in #199 now handles this case. 🙌

Issue #168, opened 2020-05-18
suggestion: change the content added to "label" attribute


jmobrien:
Right now, `read_survey()` contains a section that generate the labels from the second row of the raw downloaded response CSV, which contains a Qualtrics-constructed description of the item content.  There's some text-matching that tries to pull out the part of the second row cell believed to be most likely meaningful, but it's got some problems, and I'm thinking it might be worth just removing.  

Biggest concern is that it can lead to meaningless and ambiguous labels.  For instance, two multiple choice questions "Do you have a favorite cheese? {Yes, it's [text box entry] / No}" and "Do you live in the United States {Yes / No, I live in [text box entry]}" would both produce 2 vars: one for the choice, and one for anything written in the text box.  

Although the description cell for each variable would be different, the label-generation code would just pull out 2 identical end parts for each question's variables: "Selected Choice", and "Text Entry", with results that are then ambiguous.  Also, even if there was only one of the two items in the survey, the label still doesn't make clear what the item actually is.  I've been encountering both annoyances lot lately while trying to match variables across multiple long surveys. I eventually just gave up on our labels and pulled out the second row data myself.  

(There are sometimes other problems, too, because the code is really just implementing some heuristics built around where hyphens are found in the description cell, and it's not always perfect.)

There's clearly some thought put into this section, and I could see it making sense for certain use cases, e.g. where individual choices have clear labels, but the overall prompts are too long to be a useful label.  Still, with problems even with simple questions like MC + text box, I suspect most users will more often see the drawbacks.  I'm wondering maybe if it should be left in, but change the default behavior to use the entire cell for label-generation.

juliasilge:
Related to #132.

I would love to see some more discussion around this. I will be honest that I **never** use those labels and do not find them useful; the code to create those labels predates my work on this package. In my opinion, it is more useful to have the question text and other info in another dataframe and use joins to get that information back to the correct columns for specific analysis work.

Personally I would be pretty happy to see the labels go aways but I am not sure how folks are depending on those labels and what a good way to serve that need might be.

jmobrien:
Also related to #146, then, specifically [here](https://github.com/ropensci/qualtRics/pull/146#issuecomment-614717880).  Question text could go in the column maps, and I tend to think it should regardless of what we decide here.

[...]

PR #199, opened 2020-12-09
Alternative approach to column mapping

jmobrien:
So, looking back at #146, I realized the work I was doing was kind of a kluge, trying to build on the existing functionality that all felt really dated and that both I and others (e.g. #192) are working around to try and implement some good ideas.  So, I'm submitting this pretty complete revision to `read_survey()` that I think represents a big step forward.  This implements all the proposed functionality from #146 in, I think a much better way, along with addressing a few other things besides.  If used, this would make #146 obsolete and ignorable.

1. Column maps are now generated from the response data metadata, done cleanly and simply w/in `read_survey()`.  Now works with both the current and legacy format (FYI legacy is disappearing March 2021).
2. Generated column maps provide: 
       1. Attachable column map as data frame attribute. This is controlled by param `add_column_map`, default `TRUE`.
       2. Content for the `sjlabelled`-style label attributes. This is now controlled by new param `add_var_labels,` with default `TRUE`, addressing #134.  Label content was changed to include the entire variable description rather than the often-ambiguous subquestions, addressing #168.  The main/subquestion separation is still maintained in the column map, with a slightly cleaner implementation to generate that separation.
       3. QID-based names that also incorporate choice ID to preserve uniqueness (maintaining improvements provided by #182)
3. Simplified workflow now handles 0-response downloads the same as any other.  @lyh970817 was absolutely right to try to target the previous version's strange handling of 0-response situations in #192, but this just circumvents the issue entirely.
4. General restructuring to use "tidy" tools and create an organized workflow that should be easier to maintain.  
5. Includes the `extract_colmap()` function that outputs the column map attribute to be saved as its own object.
6. Tests added for examining the generated column maps extracted by `extract_colmap()`, implicitly testing `read_survey()` as well.

`read_survey()` is really the likely to represent the core of data processing that `qualtRics` does for most users, so hopefully this will be a useful change for everyone.  

Still maybe needs tests for `import_id = TRUE`, but that really seems like a minor revision for the current tests for `fetch_survey()`, and those depend on qualtrics surveys that aren't mine, so I thought I'd hold off on that.


jmobrien:
One FYI - An API token of mine got pushed somehow to one of the commits in a fixture.  Still not sure quite what happened there; the token appeared properly masked on the local, which was supposedly up-to-date with the remote.  Regardless, that token has already been discarded, and the current version should be properly masked.

jmobrien:
Made a few fixes to the documentation/imports/tests so that now all the checks pass.  Also see PR #200, which was a source for me of some problems with getting the tests to work appropriately.  

One other thing--I did take advantage of multi-column output from `mutate` ([here](https://github.com/ropensci/qualtRics/pull/199/files#diff-79bb8b511bf5267234e54ba785119e94739b76559aa1fb24e6b4c100f66e70d6R163-R167) and [here](https://github.com/ropensci/qualtRics/pull/199/files#diff-79bb8b511bf5267234e54ba785119e94739b76559aa1fb24e6b4c100f66e70d6R191-R202)), which means that this relies on dplyr >= 1.0.  That could be changed, but the code's a lot simpler this way.

jmobrien:
Okay, just for thoroughness I'm going to use bfg to remove my url and (expired) token.  Converting this to draft while I get that done.

jmobrien:
Didn't remove things but there's really nothing there that's actually sensitive, so this is ready for review whenever.

juliasilge:
Thank you so much for all this thoughtful work @jmobrien! 🥳  

It looks really fantastic and solves a significant challenge so many users have been having. I added a bit to the docs for `extract_colmap()` for folks to see how to use it; it worked really well on all the surveys I tried it out with.

PR #200, opened 2020-12-14
Small change to helper-qualtrics.R function for better VCR setup

jmobrien:
Current helper function for VCR setup masks the branded qualtrics URL as `<<<my_base_url>>>`.  The tests, however, expect to see `t.qualtrics.com` as part of the uri in the recorded api call.  After using that helper function, I was seeing tests fail due to mismatched uri, when everything else was good, so changing to the latter for masking should hopefully clear that up.


jmobrien:
Okay, it appears I'm not understanding this correctly.  Closing this PR.

juliasilge:
I will be honest that this is still not working entirely correctly for me either; I am having to do some manual replacing of keys and such, which I discussed in ropensci/vcr#167. Not sure what the solution is!

jmobrien:
Okay, I tracked some stuff down.  The devtools::test() framework creates a
new environment for testing, and doesn't reimport the .Renviron file where
the API key and URL are stored.  So, when `vcr::vcr_configure()` is called
within a testing run, `Sys.getenv('QUALTRICS_API_KEY')` returns `""`.
Hence, no masking.

A second problem is that the environmental variables for QUALTRICS_API_KEY
and QUALTRICS_BASE_URL are repeatedly changed throughout the testing
framework, so that even if they are right at the start of testing they're
likely something else by the time they get to a new testing file.

Adding `readRenviron("~/.Renviron")` to the first line of the file
`helper-qualtrics.R` causes the masking configuration to work
properly--provided that you call `readRenviron("~/.Renviron")` again with
the new testing file, commenting out the call to
`qualtrics_api_credentials`.

Except there's still a problem: the masking is wrong.  It masks the url as
"<<<my_base_url>>>` rather than as `t.qualtrics.com` used in all the
existing fixtures.

One might think that the mask configuration to this would fix it:

filter_sensitive_data = list("<<<my_api_key>>>" =
> Sys.getenv('QUALTRICS_API_KEY'),
>                                "t.qualtrics.com" =
> Sys.getenv('QUALTRICS_BASE_URL'))
>

And it does in a sense--it causes the new fixtures to be written
correctly.  But, it then causes ALL the other tests to fail, for reasons I
don't fully understand.

That's all I have for now. Not sure where to proceed from here.  Adding
`readRenviron("~/.Renviron")` to `helper.qualtrics.R` seems like it could
be the correct next step at least, but I wasn't clear whether that would be
kosher from a testing perspective so I haven't made a PR for that myself.

jmobrien:
I made a few changes that I think might help:

 - Loads the .Renviron before testing, so new tests work
 - Drops the masking of each person's URL.  
 - Instead, sets it to a default of "www.qualtrics.com", which should work for everyone.  
 - Adjusts the tests and fixtures to suit.  

Basically there were problems with the "t.qualtrics.com" approach - if it was wasn't set up as the mask, then new fixtures would use the wrong masking and later wouldn't match with "t.qualtrics.com" in the tests.  But if it WAS used as the mask, then vcr would convert it BACK to the real URL when it loaded the cassette, at which point it ALSO wouldn't match with "t.qualtrics.com." 

This approach just uses a non-secret URL across the board.  

New tests being written still need to comment out the call to `qualtrics_api_credentials()` when being written, as well add some lines, as follows:

```{r}
# qualtrics_api_credentials(api_key = "1234", base_url = "www.qualtrics.com")
readRenviron("~/.Renviron") # Calls back up the API key and URL likely overwritten by previous tests
Sys.setenv("QUALTRICS_BASE_URL" = "www.qualtrics.com") # Sets base URL to the default
```

Still feels clunky but it works.

tests/testthat/test-all-surveys.R
```diff
@@ -20,6 +20,8 @@ test_that("all_surveys() sends the proper request to Qualtrics", {
 
 })
 
+qualtrics_api_credentials(api_key = "1234", base_url = "t.qualtrics.com")
```
juliasilge, line 23: Does this one need to change to?

juliasilge: Ah yep, I get what's happening there. 👍

juliasilge:
OK @jmobrien I appreciate this a lot; let's give this approach a try.

PR #203, opened 2021-01-11
new function fetch_description()

jmobrien:
Adds a new "fetch" function `fetch_description()`, which is basically an updated `metadata()` using the new v3 Get Survey endpoint.  Name follows recent trends while removing the ambiguity between what qualtrics calls "metadata" and the broader survey description.

`legacy = TRUE` calls the older `metadata()` function for now, and passes any arguments (i.e. questions).  

Small change to older metadata() function to simplify and make consistent with the new approach: now no longer need a named boolean list of elements to return, can just pass a character vector with what you want.  Old approach still works with a message.

A few things remain: survey_questions() still depends on the older metadata(), as does the internal infer_data_types() called when `fetch_survey` uses `label = TRUE` (though IMO that whole relabeling system is problematic and should be deprecated). Also, some things like response counts and embedded data aren't in the new endpoint, but having worked with them, they don't always match up as expected, which is probably why they were removed.  

So, there's more to do.  But, generally, this should set us on the path to being able to work with the new metadata consistently, which is the last major piece in removing some of the ways that response data and metadata sometimes don't line up.  

Major hat tip to @lyh970817 on this, who discussed this with me somewhat, and whose related work across the two Get Survey endpoints was a key referent.


juliasilge:
@jmobrien thank you so much for all of your work! 

Do you see how there are so many commits in this PR? Instead of creating a PR in this way, branched off of master a long time ago with conflicts, can you [create a fork of this repo](https://usethis.r-lib.org/articles/articles/pr-functions.html) with `usethis::create_from_github()`, and then initiate a new branch with `usethis::pr_init()`? You might find [this chapter of Happy Git with R](https://happygitwithr.com/common-remote-setups.html) helpful; you want to head for "fork-and-clone".

jmobrien:
Thanks.  I was aware that the commits had gotten a bit out of hand.  I'm self-taught on Git and haven't had any curated input on the proper flows.  I tried to rebase and pre-resolve here so it wouldn't get too out of hand, but it obviously didn't really produce the intended result.  I'll read over those resources and re-submit sometime soon.

juliasilge:
That would be so great! 💯  Learning to use git can be really **a lot** but those [PR helpers in the usethis](https://usethis.r-lib.org/articles/articles/pr-functions.html) package are incredibly helpful.

PR #207, opened 2021-03-08
New function fetch_description(), replacing metadata() for v3 Get Survey endpoint

jmobrien:
Trying this again--here's a PR for a new function fetch_description() that queries the new v3 Get Survey endpoint, replacing the v2 endpoint queried by the existing `metadata()`

To simplify things, it now returns all elements with some modest processing on a few parts.  Parameter `elements` (analogous to the old `get` param from `metadata()`) now takes a character vector rather than `get`'s more complex named logical list.  Old approach of passing lists is preserved, with a message.

Older endpoint can be called by setting legacy=TRUE, which passes everything to `metadata()`  Passing a named list to `elements` here (and with `get` when calling `metadata()` directly) maintains the original functionality: "metadata", "questions", and "responsecounts" will always be included unless explicitly set to FALSE in a named list.  (If using the new character vector approach, though, you just get what you requested.)

Do note: what's available from v3 is not the same as what came from v2.  In particular, response counts are now gone--but I'm hearing they may have been somewhat unreliable anyway.


jmobrien:
Seems like the failure relates to package httr not being currently available in linux?  Not clear if that's directly related to this PR but if it needs something let me know.

juliasilge:
This looks fantastic @jmobrien and worked really well on my surveys. 🙌 

One thing to keep in mind for the future for the `metadata()` function, or for future similar PRs, is to consider soft deprecating the older function like `metadata()` instead of building code to support both. I like to use [lifecycle](https://github.com/r-lib/lifecycle) for that, and you start with just a soft deprecation. It is tough to keep supporting all older functions forever (code maintenance EXPLOSION) so it's something to keep in mind.

juliasilge (approved):

jmobrien:
Thanks, good to know on best practices for lifecycle!  Along with just not being sure of my place/how best to do deprecation, I held off partly b/c v3 drops some things that were in v2, like response counts and separate listings for embedded data.  Without a sense of how much the userbase wants those, I went with something like the `legacy` switch used in `read_survey` to preserve their presence more explicitly.  

Also, it's worth mentioning we have at least one helper function `survey_questions` that's completely built around the v2 endpoint and `metadata`, which would need a complete redesign for v3.  I find `survey_questions` confusing and don't use it (I think it might include trashed/non-displayed questions, and maybe not in order?  Column maps seem more useful overall), but others might see it differently.  

Not a defense of the choices for this PR, moreso I'm just trying to acknowledge there were some questions on direction/functionality I wasn't sure about, things possibly important enough to warrant further consideration.

juliasilge:
Yeah totally, makes sense 👍

Issue #219, opened 2021-05-19
Feature: Download distribution history


Rory-OG:
Can we build a function to pull the distribution history? This aligns with the ability to know who bounced. In some instances, it's easiest to gauge true participation rates by filtering out the bounces. Additionally, the fact that I have to pull the history to get the participant list is why pulling the history would be helpful as the history pulls everyone the survey went out to.

dsen6644:
Related to #206

Rory-OG:
Is this function usable in the dev version?

juliasilge:
It's not yet here in the main branch @Rory-OG but if you read through that other thread, I believe you can try out what @dsen6644 has [built out in his fork](https://github.com/dsen6644/qualtRics). Definitely would love to get that figured out and ready for general use!

chrisumphlett:
IMHO with the minor changes I suggested at the end of the #206 thread it should be ready to be merged in. I'm using @dsen6644's fork (with the changes) as part of my regular data collection.

juliasilge:
That is great to hear!! 😄 @dsen6644 would you be interested in opening a PR?

chrisumphlett:
I made a PR to @dsen6644's fork with the changes I've made. #217 comes into play. I have my method which might not be what you'd ultimately want included, but it's in my branch.

dsen6644:
Sorry for the late response, I think for this to go into production we need two things.

1. A warning/message specific to the `fetch_distribution_history` function that informs those with basic licenses that this function only works for XM directory users. Basic license users will receive a 404 error when calling the function.
2. We need a `fetch_distribution_history` test and its respective .yml file. Since I have a basic license I won't be able to run the test. For those who can, review #201 for workflow.

juliasilge:
No worries at all about timing! @dsen6644 if you can open a PR with what you have already done and outline what you believe still needs to be done, we can get the other necessary things added in.

dsen6644:
PR has been submitted #221

juliasilge:
After merging in #239, I believe we still need this:

>     1. A warning/message specific to the `fetch_distribution_history` function that informs those with basic licenses that this function only works for XM directory users. Basic license users will receive a 404 error when calling the function.
> 

We also should add a note to the documentation for the relevant functions that they require a certain kind of license.

Anything else still outstanding for this issue?

chrisumphlett:
[...]

Issue #206, opened 2021-02-24
Feature request - detailed list of who bounced/opted out/etc


chrisumphlett:
I just used `fetch_distributions` and `fetch_mailinglist` for the first time.  I'm interested in getting invitee-level info on the things that are returned from `fetch_distributions` like `stats_bounced` and `stats_blocked`. I do not find Qualtrics documentation very easy to read, since mailinglist returns an element for each invitee I'm thinking that's probably where to look (even though distribution-level stats are very useful for me right now).

[This link](https://api.qualtrics.com/api-reference/reference/mailingListContacts.json/paths/~1directories~1%7BdirectoryId%7D~1mailinglists~1%7BmailingListId%7D~1bouncedContacts/get) is for "List Bounced Contacts in a Mailing List." I'd think there would be something where you could get a boolean flag for each of the stats from `fetch_distributions` but I didn't see that.

[There's a distribution history endpoint](https://api.qualtrics.com/api-reference/reference/distributions.json/paths/~1distributions~1%7BdistributionId%7D~1history/get) with contact level data but this returns sent/open/completed but not bounced/blocked.

dsen6644:
`all_mailinglists`, `fetch_mailinglist`, and `fetch_distributions` were developed first because those endpoints are supported for basic level licenses (i.e research core) and would work for all users.

It would be great to expand their functionality to include distribution and mailing list history, distribution links and expiration dates, etc. but would require developers to have access to a license with XM Directory. 

This is something I have wanted to do for a while, but unfortunately, my account does not have XM Directory. As a result, when I try to call those APIs endpoints linked above I get 404 errors.

@juliasilge, you mentioned we have a test account that we use for development, does it have XM Directory? If not, do you think we should contact Qualtrics and see if they have any interest in upgrading our dev account?

chrisumphlett:
I am told that we have XM Directory at my org so I could help with that testing/development.

chrisumphlett:
@dsen6644 If I help with the testing (and willing to help with the dev too, I would just want a bit of a nudge to make it a bit easier), is that something we could work on sooner rather than later?

dsen6644:
[...]

PR #221, opened 2021-06-15
Add `fetch_distribution_history` and `list_distribution_links`

dsen6644:
Items still left to do

- [ ] Create a warning/message specific to the fetch_distribution_history function that informs those with basic licenses that this function only works for XM directory users. Basic license users will receive a 404 error when calling the function.
- [ ] Tests need cached .yml file in fixtures.

Also noting, I did not use @chrisumphlett error handling. I think that does need to be addressed at a higher level, especially since it's being seen across multiple functions.

Addresses to #206 and #219.


chrisumphlett:
FWIW when I ran my fork most recently, with the error handling, I never noticed it actually having errors (I didn't look back through the entire console, it was taking 500-700 iterations). When I was using this endpoint a few months ago I was hitting an error every time when doing 100-200 iterations

juliasilge:
So I am having a hard time testing this PR out 😔 because I am not finding out how to "fake" or other mock up distribution data, and I don't have any in my test account. This also means I can't even generate the 404 to test against. I think we are going to need some external input or help to finish this PR up.

What do we need?

- [ ] Someone with a basic license who has used distributions to test _trying_ to get the history and failing so we can generate a nice error
- [ ] Someone with the fancy license who has used distributions who can help get the tests written

Is this right? Anything else that we know of?

R/fetch_distribution_history.R
```diff
@@ -15,7 +16,10 @@
 #'
 #' surveys <- all_surveys()
 #' distributions <- fetch_distributions(surveys$id[1])
-#' distribution_history <- fetch_distribution_history(distributions$id[1])
+#' distribution_history <- fetch_distribution_history(
+#'     distributions$id[1],
+#'     surveyID = surveys$id[1]
+#' )
```
juliasilge, line 22: Is this correct? The `surveyID` arg is required, right? (I will fix them to match style-wise in the next commit, with both not needing names.)

dsen6644: Correct, surveyID is required if you want emails associated with the distribution list and not just contact ids. The survey id is necessary for `list_distribution_links` embedded in `fetch_distribution_history`. Feel free to make any style changes you see necessary.

@chrisumphlett may be able to help with the formal testing given that he has an XM directory license.

juliasilge:
Closed in #239

Thank you so much @dsen6644! I attributed your work here in the NEWS 🙌

PR #225, opened 2021-08-09
simpler & more extensible way to generate endpoint URLS, closes #216

### LINKED ISSUE #216: Refactor URL creation functions
We have built up quite a number of functions to build URLs (`create_distributions_url()`, `create_survey_url()`, etc etc etc)) and at this point it would make more sense to have one function for URL creation with a `switch` statement or similar.

jmobrien:
Thought I'd just tackle this quickly since I had a plan, so others can be free do do more interesting development.  

URL creation now depends on a single utility function `generate_url()`, which takes params `query` and, where appropriate, `id`.  

`query` is the thing being called, generally named the same as the associated (primary) function w/o the underscore, like "fetchdescription", "fetchsurvey", etc.  

`id` is that of the survey/mailinglist, etc.  ID-less functions like `all_surveys()` ignore this.

Inside the function is a list of strings that define how to build a URL for each type, using a notation passed to the base R function `sprintf()`.  To extend `generate_url` for new uses, all you need to do is add another line to this list with a new specification.  

Submitted, but marked for draft so I could mention the idea of using `glue()` rather than `sprintf()`.  While that would introduce a new (tidyverse) dependency the package doesn't currently need, it would allow a more readable form of the URL definitions list.  The various `%s`'s  passed to `sprintf()` could be replaced with explicitly named alternatives like `{root_url}`, `{id}`, etc.  

Didn't regenerate documentation as I don't have the development version of roxygen2.


jmobrien:
Update: upon realizing PR #221 already was now using multiple IDs in url generation, it was clear the above was already too limited.  So, instead I updated to use the `glue::glue` approach, which gives us the needed flexibility going forward.  

Also, shifting to a full PR now so this can move forward if this suffices for everyone.

jmobrien:
FYI due to the change all params passed to `generate_url` (other than to `query`) must be explicitly named, using whatever is entered into it's associated template string.

juliasilge:

Issue #223, opened 2021-07-29
Feature: Include Embedded Data in Survey Questions


drsjp31:
I am looking to fetch just embedded data from a survey. fetch_survey does included the embedded data when it's pulled as a whole, but it doesn't recognize the embedded data column names in the include_questions argument. I tried the code below (where survey_ID is a string) and received Error in qualtrics_response_codes(res) : Qualtrics API raised a bad request (400) error. If I remove the include_questions argument, it runs properly and EventCode is included as a column in the data output.

fetch_survey(surveyID = survey_ID, 
             force_request = T, 
             include_display_order = F, 
             label = F, 
             convert = F,
             include_questions = c("EventCode"))

Using the metadata function, I can see EventCode as an embedded data variable, but when I run survey_questions(survey_ID) it only returns the actual questions in the survey, which I assume is linked to the include_questions argument. I'm looking for a feature that would allow include_questions to also refer to the embedded data columns so I can extract what I want versus having to pull the entire data set first.

juliasilge:
I just spent a little bit of time on this and it looks to me on an initial glance that the Qualtrics API won't send back just embedded data at the endpoint we are using; I believe you'll need to query the API (via this R package) with at least one survey question.

I'll keep this open in case we find a different way to approach the problem. Thanks for the feedback! 🙌

jmobrien:
@juliasilge fiddling around, I stumbled on using ` include_questions = character()` to leave out all questions.  So, that works, but I thought passing `NA` made more sense for the UI, so I did that real quick in draft PR #226.  

Hope that helps, but it still needs tests, and I'm about to dive back into some other work for a while (days, maybe). So, if anyone wants to add the tests to finish that PR I'd be delighted!  Otherwise I'll come back to it when I can.

@drsjp31, for now, I think you can do the below to leave off all questions, while just keeping (all) embedded data vars:

``` r
fetch_survey(
   surveyID = survey_ID,
   force_request = T,
   include_display_order = F,
   label = F,
   convert = F,
   include_questions = character()
)
```

juliasilge:
Ah nice, that does work for me! Only non-question columns:

``` r
library(qualtRics)
rs <- fetch_survey("SV_56icaa9YAafpAqx", include_questions = character())
#>   |                                                                              |                                                                      |   0%  |                                                                              |======================================================================| 100%
#> 
#> ── Column specification ────────────────────────────────────────────────────────
#> cols(
[...]

PR #226, opened 2021-08-09
specify questions, embedded, & metadata, or drop w/NA, closing #223

jmobrien:
Adds `include_embedded` and `include_metadata` to fetch_survey().  

Also, can now specify NA in any param to exclude all variables under that category.  

Still needs tests, but I'm out of time for now.


jmobrien:
Discarding this as this was folded in to #263.

PR #227, opened 2021-08-26
fetch_survey() by survey_name

markjrieke:
Added param survey_name to fetch_survey() to allow for users to fetch_survey by the Qualtrics UI survey name, rather than the surveyID. I added a check to ensure that either surveyID or survey_name is provided, and another check for duplicate names (I'm not sure if this is allowable in the platform, but I added here just in case). surveyID is still an accepted parameter, but with the default changed to NULL. I ran a test()/test_file() call & came back w/no warnings or errors, but would be happy to make changes as needed!

This is a more fleshed-out version of a helper function I would load into my environment every time I loaded the qualtRics package - I figure it may be more beneficial to bring into the package itself.

### DIFF
--- a/R/fetch_survey.R
+++ b/R/fetch_survey.R
@@ -4,6 +4,8 @@
 #'
 #' @param surveyID String. Unique ID for the survey you want to download.
 #' Returned as \code{id} by the \link[qualtRics]{all_surveys} function.
+#' @param survey_name String. Name of the survey you want to download as it
+#' appears in the Qualtrics UI.
 #' @param last_response Deprecated.
 #' @param start_date String. Filter to only exports responses recorded after the
 #' specified date. Accepts dates as character strings in format "YYYY-MM-DD".
@@ -101,7 +103,8 @@
 #'
 #' }
 #'
-fetch_survey <- function(surveyID,
+fetch_survey <- function(surveyID = NULL,
+                         survey_name = NULL,
                          last_response = deprecated(),
                          start_date = NULL,
                          end_date = NULL,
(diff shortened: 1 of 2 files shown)


juliasilge:
Thank you so much for this contribution @markjrieke! 🙌 

I spent a little bit of time looking at this and how Qualtrics handles the [concepts of names for surveys](https://www.qualtrics.com/support/survey-platform/survey-module/renaming-your-survey/), and I think it might be too ambiguous and confusing to have survey name as an argument for the package's main survey fetching function. It sounds like your surveys may have unique or straightforward names, but in orgs where I have worked, that has typically not been the case. I think this functionality is likely to open up problems and confusion, rather than be a shortcut in general.

It sounds like you are in a situation where you know the name of your survey and want to use it for `fetch_survey()`. What if, instead of putting the call to `all_surveys()` inside of `fetch_survey()`, you created a helper function that would find the `surveyID` for a given name? A function that does this, with some helpful error checking:

```r
surveyID <-
        all_surveys() %>%
        dplyr::filter(name == survey_name) %>%
        dplyr::pull(id)
```

My first preference would be that it would take a dataframe from `all_surveys()` as an argument (rather than calling `all_surveys()` internally) but let me know what you think.

markjrieke:
Hey Julia, thank you for the feedback! You're definitely correct regarding my co's use case - all have unique names & I've found it slightly less time consuming to write a quick extension of `fetch_survey()`

Regarding your second point, that would allow for piping into the existing `fetch_survey()` , here's a quick ex:

```R
fetch_name <- function(x, survey_name) {

  x %>%
    dplyr::filter(name == survey_name) %>%
    dplyr::pull(id)

}

all_surveys() %>%
  fetch_name("Unique Survey Name") %>%
  fetch_survey()

```

For my use case, I'm covered by the local functions I have, but if this is something that would be useful for the overall package, let me know! Either way, it sounds like this specific PR for modifying `fetch_survey()` can be closed out.

juliasilge:
Yeah, that fluent piping is exactly what I was thinking about, and having it as a separate helper function lets us handle errors in _that_ function rather than make `fetch_survey()` yet more complex and try to depend on naming characteristics that Qualtrics doesn't promise us. I think that kind of helper function could be super useful for a lot of people and if you would be open to working on it, that would be most welcome! 🙌 

It probably would be easiest to close this PR and open a new one rather than change this one.

markjrieke:
Gotcha - can do!

PR #230, opened 2021-09-13
add new function fetch_id() to return a surveyID for a unique survey_name

markjrieke:
Added a new function `fetch_id()` that returns a `surveyID` for a unique `survey_name` as it appears in the Qualtrics UI (based on the conversation in [PR 227](https://github.com/ropensci/qualtRics/pull/227)). `fetch_id()` checks that the `survey_name` is unique, and will throw an error if a non-unique `survey_name` is passed. 

I added a test for `fetch_id()`. It's not interacting with the API and to keep separate from `all_surveys()`, the tests are based on a toy dataframe that mimics what `all_surveys()` currently returns. That may be superfluous (since it's essentially testing error messages & that `dplyr::filter` and `dplyr::pull` still function as expected), but I figured it'd be better to have test coverage over the new function than not. 

If there are any comments/thoughts/edits, let me know!


markjrieke:
Ah it looks like the checks are failing due to the lines `qualtRics::fetch_id` in the test file. The changes pass all tests when I run `test_local` but similarly fails when I run `check()` (should have run that prior, but learning as I go here!). I'll get that sorted out locally then push

juliasilge:
Thank you so much for this contribution @markjrieke! 🙌

PR #239, opened 2021-10-28
Add fetch_distribution_history

chrisumphlett:
Let's break this into a few, smaller pieces. Ultimately this function isn't that useful without the distribution links in my opinion. But I'd like to get this working, and address issues with it, before adding the links.

This works. Getting it to work on a distribution with a lot of invitations, and potentially doing error-handling, can be handled later (I tried twice, it got through it once and failed the other time). It will consistently work on a small distribution.

I do want to bring attention to a 400 error I received:

> "{"meta":{"requestId":"6d98dd48-0cd6-4a1e-994b-dccd785fc526","httpStatus":"400 - Bad Request","error":{"errorCode":"CP_TD_GER_NP","errorMessage":"The email-distribution(REMOVED) has no recipients and no parent. This distribution type may be unsupported."}}}"

This distribution was some kind of test someone did. `requestType`, a column that is returned by `fetch_distributions()`, has a value of "TestEmail" whereas it is normally "Invite". Some error handling for this should be added, otherwise you get this which isn't particularly helpful:

```
> dist_hist <- fetch_distribution_history("REMOVED")
 Error: Qualtrics API raised a bad request (400) error - Please report this on
https://github.com/ropensci/qualtRics/issues
Run `rlang::last_error()` to see where the error occurred. 
> rlang::last_error()
<error/rlang_error>
Qualtrics API raised a bad request (400) error - Please report this on
https://github.com/ropensci/qualtRics/issues
Backtrace:
 1. global::fetch_distribution_history("REMOVED")
 2. qualtRics::qualtrics_api_request("GET", url = fetch_url)
 3. qualtRics::qualtrics_response_codes(res) R/utils.R:333:2
Run `rlang::last_trace()` to see the full context.
> rlang::last_trace()
<error/rlang_error>
Qualtrics API raised a bad request (400) error - Please report this on
https://github.com/ropensci/qualtRics/issues
Backtrace:
    x
 1. \-global::fetch_distribution_history("REMOVED")
 2.   \-qualtRics::qualtrics_api_request("GET", url = fetch_url)
 3.     \-qualtRics::qualtrics_response_codes(res) R/utils.R:333:2
```

For myself, I'd just know what I'm doing and document this in the function notes. I assume that ropensci would want to be a bit more formal in the approach ;)

To summarize, I'm proposing to make changes/PR in this order:

1. Merge this PR
2. Someone else address the API error (or at least help me with how it should be done)
3. Address transient API 500 errors that I get when running this on larger distributions. Need to re-try if that happens, rather than stop the whole thing.
4. Create `fetch_distribution_links`
5. Add call to `fetch_distribution_links` to `fetch_distribution_history` as @dsen6644 originally did
6. Error-handling if necessary if getting 500 errors in `fetch_distribution_links`

Regarding (3) above, As a reminder of something we discovered before... for some reason the Qualtrics pagination is inconsistent. Sometimes I'd get 17 rows, sometimes 88. It results in having to make a lot more calls than it should be (In my case, sometimes 700+ for distributions of 30-40k emails). The version I ran locally would display the iterations of the pages with a message. That's probably too chatty as a default but it might be a nice to have for others working with large distributions who are worrying about how slow it is.


juliasilge:
This is looking great @chrisumphlett and I think that breaking this into pieces sounds good.

- Would you mind opening an issue and making a list of what else needs to be done so we can track it there? Maybe just another comment in #219?
- This does need some tests/test fixtures for the new functions. I can't remember if you have contributed tests here in this repo before; it is a bit of a pain but not too onerous, I think. The [rOpenSci book on HTTP testing](https://books.ropensci.org/http-testing/) is the best place to look, and we use vcr and webmockr. I just have to look up the process every time I need to add new tests. [This PR](https://github.com/ropensci/qualtRics/pull/207) included a test fixture if you want to see.

chrisumphlett:
I'm working on the tests. I based if off `fetch_distributions()`. I'm able to run the test for that function. But the test for `fetch_distribution_history()` fails with what seems like an API error. I'm not sure why. My hypothesis was that it could be because of this line in the yaml file being all lowercase? The yaml for `fetch_distributions()` maintained case sensitivity.

Despite (quickly) reading the first 6 chapters of the http testing book I don't know enough to start troubleshooting or understand the order of what's happening here.

chrisumphlett:
The first time I run the test (after deleting the fixture yaml) I get this error: `Error: Qualtrics API complains that the requested resource cannot be found (404 error).
Please check if you are using the correct survey ID.`

This doesn't make sense, there's not a survey ID involved. When I run it a subsequent time I get the error I shared before.

404 errors happen with a manual call when the distribution ID is the wrong length:

* `[1] "{\"meta\":{\"requestId\":\"863d8f41-cd55-43a4-85cf-e01d8d4e83bf\",\"httpStatus\":\"404 - Not Found\",\"error\":{\"errorCode\":\"CP_TD_GED_NF\",\"errorMessage\":\"The email-distribution(EMD_rmjB8Gfpjlki6m) was not found.\"}}}"`
* `[1] "{\"meta\":{\"requestId\":\"705e648d-1394-47c2-b604-5e2a72b968cb\",\"httpStatus\":\"404 - Not Found\",\"error\":{\"errorCode\":\"CP_TD_GED_NF\",\"errorMessage\":\"The email-distribution(EMD_rmj) was not found.\"}}}"`

400 errors result when the case is changed: `[1] "{\"meta\":{\"requestId\":\"de47afcc-94e0-43c1-96db-dd274e80d9e1\",\"httpStatus\":\"400 - Bad Request\",\"error\":{\"errorCode\":\"CP_TD_GR_UT\",\"errorMessage\":\"The type of distribution(emd_rmjb8gfpjlki6ms) is unknown or unsupported.\"}}}"`

The id is stored all lowercase in the yaml file. 

This might all be red herrings but it makes it difficult for me to understand what the actual problem is. That said, I did realize that if I delete the `fetch_distributions()` yaml and re-run the test with my own survey id, it also fails in the same way. So the problem probably lies somewhere with me.

juliasilge:
OK, so it is a bit part to debug remotely, as you can imagine but here are some ideas:

- The vcr setup file sets some "holder" API credentials. Is this file being executed when you are trying to _record_ your fixture? You need to NOT have these credentials in place but instead your real credentials in order to record the API call.
- Can you run `fetch_distribution_history("EMD_RmjB....")` interactively? Can you run it in the test file outside of recording a fixture?
- What about one level further down, the `qualtrics:::qualtrics_api_request("GET", url = "your_uri")`? Can you run that interactively? What about in the test file?

Try out some of those things and let me know what's happening. Thanks for hanging in there through this. 💪

chrisumphlett:
Below, the results of your 2nd and 3rd bullet point. I did this after running the following of course... which I don't do when running the test. The way the credentialing happens with the test is not making sense to me which doesn't help.

image

To your question about the vcr setup. I have two files, one called `helper-qualtRics.R` and one called `setup-qualtRics.R`. One of those was created by the setup process I followed at the book, one I think was either already in the repo or I created it to try and mimic a previous PR. I added stuff to the file according to the directions here: https://books.ropensci.org/http-testing/vcr.html

```
library(vcr)

# Load in credentials, if saved (for new tests):
readRenviron("~/.Renviron")

# Use default URL rather than branded one:
Sys.setenv("QUALTRICS_BASE_URL" = "www.qualtrics.com")

# Set up a fake API key if none is saved
if (Sys.getenv("QUALTRICS_API_KEY") == ""){
  Sys.setenv("QUALTRICS_API_KEY" = "1234")
}

# Store these for resetting later when needed
holder_API <- Sys.getenv("QUALTRICS_API_KEY")
holder_URL <- Sys.getenv("QUALTRICS_BASE_URL")

# NOTE: If writing new test that changes the credentials for some reason, add
# this line (uncommented) to end of the testing file:
# qualtrics_api_credentials(api_key = holder_API, base_url = holder_URL)



# Set directory and mask API token
invisible(vcr::vcr_configure(
  dir = "../fixtures",
  preserve_exact_body_bytes = TRUE,
  filter_sensitive_data =
    list(
      "<<<my_api_key>>>" = Sys.getenv('QUALTRICS_API_KEY')
    )))
```

_Is this file being executed when you are trying to record your fixture?_ I don't know. How would I know?

_You need to NOT have these credentials in place but instead your real credentials in order to record the API call._ I would think it would use mine. `Sys.getenv("QUALTRICS_API_KEY")` produces my key so I'd think that `holder_API` would have my token.

juliasilge:
OK, so we know you can run these functions if you have your auth set 👍  so we just need to get the fixtures saved from that.

I don't think you should have *both* a "helper" file plus a "setup" file; I think you should just use the one we already have in the repo (`helper-qualtRics.R`). That file should read **your** environment variables  and then do all the vcr stuff.

- What happens if you delete the new `setup-qualtRics.R` file, and try creating the fixture?
- I feel a little suspicious of line 8 here (won't this override your URL?):
https://github.com/ropensci/qualtRics/blob/6e64d2c7d49d113a86a63f9f44fd6eb9f129fc83/tests/testthat/helper-qualtRics.R#L8
What if you try commenting this out and then creating the fixture?

chrisumphlett:
Thanks @juliasilge doing both of those seems to have done the trick. I had wondered about that line too. I'll clean things up and push the changes this afternoon

chrisumphlett:
there are now changes to some files I wasn't expecting, like this one below in DESCRIPTION and the addition of a `.gitattributes`. I'm going to omit these, and just include the new test and fixture files but let me know if I should include more.
image

chrisumphlett:
ahh shucks, errors in the test with the api credentials. not quite there yet...

juliasilge:
I think you were just about there; for the testing in the package, we do need to use that default URL. It is only when you are *recording* the fixture that you need your real URL.

🤞 watching CI now 🤞

chrisumphlett:
`Error: Error: Please install xml2 package` ??

juliasilge:
I do **not** know why xml2 wasn't getting installed, as it is imported by httr. 🤷‍♀️ I might ask somebody here at RStudio, but for now we can just move forward with it in DESCRIPTION.

Do you think you can add a fixture for `list_distribution_links()`? You should be able to pull my changes without too much problem (I hope) and then push more changes afterward.

chrisumphlett:
oh yea, still need to do that. I was focused on getting the first one to work. I'll get that done soon, thanks for adding the xml package

juliasilge (approved):
OK, I think this is 👏 READY 👏  TO 👏 GO 👏 


@chrisumphlett can you make a list of what else needs to be done beyond this PR, in #219 or a new issue? We wanted to try to do more warnings in certain situations, right? Anything else?

This PR closes #221

chrisumphlett:
sure, I'll open a new issue after working with the new dev version of the package and document my problems I have with it when working with our larger distributions

PR #244, opened 2021-11-23
Small change to prevent converting literal "NA"s as missing.

jmobrien:
Set `type_convert()` to use `na = character()` to avoid converting anything further to `NA` (all suitable NAs should have been identified by the `read_csv()` call just above.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 - Add `fetch_distribution_history()` and `list_distribution_links()` for more handling of distribution data, thanks to @chrisumphlett and @dsen6644 (#221, #239)
 
+- Changed handling of literal `"NA"` text input from users so it is no longer converted to an R `NA` value thanks to @jmobrien (#244)
+
 # qualtRics 3.1.5
 
 - Add `fetch_description()` to download complete survey description metadata from v3 API endpoint (more up-to-date than older `metadata()`) thanks to @jmobrien (#207)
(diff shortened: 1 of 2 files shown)


juliasilge:
I wonder if it would be better to use the dots `...` here for all arguments passed to `type_convert()`, and then we make `...` on `fetch_survey()` for most of the arguments that pass through to `read_survey()` like `col_types`, etc.

jmobrien:
Yeah, I had similar general thoughts, but I think a fixed argument works best for this specific issue. Looks like the endpoint only ever offers up `""` for truly missing data (ignoring seen-but-missing).  Unless I'm reading this wrong, setting those  `""`'s to `NA` is already covered by the `read_csv()` call on line 115.  So, adding `na = character()` to `type_convert` just eliminates any extra missingness checks, which I think is what we want.  

Considering the potential for something like `...`, the potentially-flexible args for `type_convert` are `col_types`, `locale`, and `trim_ws`.  We have `col_types` already, but only partly support `locale` through `time_zone`, and nothing for `trim_ws`.  `...` could offer consolidation and add support for whitespace removal.  However, it's worth noting that the central role for the standalone `time_zone` arg in `fetch_survey()` is as a param for the API query.  

So, I'm not sure.  Though, I am thinking this PR is just a bugfix that can be considered separately.

juliasilge:

Issue #217, opened 2021-04-22
Implement standard error handling for Qualtrics API errors


chrisumphlett:
In my experience getting a 500 error from Qualtrics is a normal, intermittent issue. When using a function that will need to iterate many times to get its results (a lot of survey responses from `fetch_survey` or many contacts from `fetch_distribution_history`), if it gets that error one time, you have to start over. In some cases I have not been able to ever see the function finish.

I opened a ticket with Qualtrics, who gave this explanation:

> Qualtrics' servers return a 500 error when they've met such a spike in volume that they can't handle all calls being made at that moment. The good news is that this means that any calls returning a 500 error can simply be sent again without changing any parameters with the expectation that the second iteration goes through without issue. 
> 
> Qualtrics strives to maintain a 500 return rate of roughly one percent, and in the past week, calls made from your license have had a 0.99% 500 rate. Should this rate cross 4% or you find that a single call (that is, one payload and one endpoint) consistently returns 500 errors, please do reach out and let us know, but for the time being, what you're experiencing is "within spec." With the implementation of logic in your automation to retry calls that fail with a 500 status, you'll be set to go!

The easiest and best thing would probably be to have a single error-handling function and then wrap it around each call to the qualtrics API. Below is an example of how I've done this. I don't think this is exactly the right solution though. For one, this is wrapped around the `fetch_survey` function, not the call inside the function. This works for me because I'm looping through many surveys, and if one survey fails, I can just re do the fetching on the entire thing and it's not a big deal because I'm never pulling a lot of results from any one survey.

```
for (i in 1:length(active_survey_ids)) {
    # for (i in 22:22) {
    print(paste0("Getting response data for ", active_survey_names[i]))
    tryCatch({
      # fetch responses for each survey
      # This will often fail the first time. Try up to three times
      # https://stackoverflow.com/questions/28969070/assigning-a-value-in-exception-handling-in-r/28969896#28969896
  
      attempt_fetch <- 1
      while(attempt_fetch != 4){
        responses <- try(responses <- fetch_survey(surveyID = active_survey_ids[i], verbose = TRUE, force_request = TRUE,
                                                   start_date = as.character(as.Date(as_of_date) - 6), end_date = as.character(as.Date(as_of_date) + 1)) %>%
                  rename_all(.funs = tolower))
        if (class(responses)[1] == "try-error") {
          Sys.sleep(5) ## Wait 5 seconds so that transient connection issues can go away
[...]

PR #247, opened 2022-01-10
use httr::RETRY instead of httr::VERB (#217)

chrisumphlett:
This will implement consistent error-handling across all of the api call functions in the package. They will be retried up to 3 times if there is any error.

There's an optional argument `terminate_on` that allows you to specify the statuses where this would NOT happen. This is really only relevant for 50X errors. If you have the wrong credentials, we don't need to retry. I haven't used this argument before, and the documentation doesn't specify how specific it needs to be (does "400" cover 40X?). 

Here's an example, I give a bad survey Id value to `fetch_distributions()`. It repeats the request 3 times, then uses the package's existing error handling to tell me what happened.

image

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -338,8 +338,8 @@ qualtrics_api_request <- function(verb = c("GET", "POST"),
   headers <- construct_header(Sys.getenv("QUALTRICS_API_KEY"))
   # Send request to Qualtrics API
   res <- httr::RETRY(verb,
-                    url = url,
-                    httr::add_headers(headers),
+                     url = url,
+                     httr::add_headers(headers),
                     body = body,
 					times = 4
   )
--- a/R/utils.R
+++ b/R/utils.R
@@ -340,7 +340,7 @@ qualtrics_api_request <- function(verb = c("GET", "POST"),
   res <- httr::RETRY(verb,
                      url = url,
                      httr::add_headers(headers),
-                    body = body,
+                     body = body,
 					times = 4
   )
   # Check if response type is OK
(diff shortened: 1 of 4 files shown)


juliasilge:
This seems to be working great @chrisumphlett. Unfortunately, it makes the tests take too long for CRAN. The specific tests that are now problematic are:

- `"it should throw an error after certain 400 and 500 status codes"`
- `"all_surveys() throws an error when URL & key is bad"`

My inclination is to use `testthat::skip_on_cran()` for these, unless there is some other way to address this. Thoughts?

Can you also update the NEWS file?

juliasilge:
I tried changing `pause_cap` to 10, and then 5, and those tests still take too long for CRAN (still almost a minute each). Can you add `skip_on_cran()` to those two tests? I tried to get your PR locally to do it myself, but I think you may have forked off an earlier version of the main branch; I am having trouble getting your PR.

R/utils.R
```diff
@@ -337,10 +337,11 @@ qualtrics_api_request <- function(verb = c("GET", "POST"),
   # Construct header
   headers <- construct_header(Sys.getenv("QUALTRICS_API_KEY"))
   # Send request to Qualtrics API
-  res <- httr::VERB(verb,
+  res <- httr::RETRY(verb,
                     url = url,
                     httr::add_headers(headers),
```
juliasilge, line 342:
```suggestion
                     url = url,
                     httr::add_headers(headers),
```

R/utils.R
```diff
@@ -337,10 +337,11 @@ qualtrics_api_request <- function(verb = c("GET", "POST"),
...
-                    body = body
+                    body = body,
```
juliasilge, line 343:
```suggestion
                     body = body,
```

juliasilge (approved):
Thank you so much @chrisumphlett! 🚀

Issue #249, opened 2022-01-13
Qualtrics update on 6/1/2022 will break fetch_distributions()


chrisumphlett:
Our org got an email about two changes to the List Distributions endpoint (https://github.com/ropensci/qualtRics/blob/master/R/fetch_distributions.R) which is used in the `fetch_distributions()` function in `qualtRics`.

The first change is to how pagination works, and I believe that no change will be required because we already use `nextPage`.

The second change will remove some fields that are currently mapped to columns in the resulting data.frame. I have not personally used the affected columns and do not believe it's necessary right now to make the connections to other endpoints that the email describes.

Is there a standard way to manage something like this in GH, where the work can be done in advance but it shouldn't be merged in?

Columns changing: 
![image](https://user-images.githubusercontent.com/7210482/149378956-6399f304-ac2e-4d29-bff2-bbef18156b22.png)

The entire email:
![image](https://user-images.githubusercontent.com/7210482/149379107-55110891-17a0-45b4-aa1b-c698763d0c65.png)

juliasilge:
Thank you for sharing this! I think the best thing to do is to create a [draft PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) with the needed changes and leave it as draft until we are really there at needing to be merged in.

juliasilge:
Closed in #250

PR #250, opened 2022-01-14
on 6/1/22 qualtrics will change the api, remove 2 cols and add 1

chrisumphlett:
#249 
This shouldn't be merged until June


juliasilge:
Are you able to update the test objects @chrisumphlett?

```
── Failure (test-fetch-distributions.R:32:3): fetch_distributions returns a tbl_df with expected column names and types ──
x$message_messageText has type 'NULL', not 'character'.
── Failure (test-fetch-distributions.R:36:3): fetch_distributions returns a tbl_df with expected column names and types ──
x$embeddedData has type 'NULL', not 'character'.
```

juliasilge (approved):
Thank you so much @chrisumphlett! 🙌

Issue #248, opened 2022-01-12
Update documentation, console messaging, and/or README for new functions and error handling


chrisumphlett:
1. On the README, should the newer functions for distributions show up in the **Access your Qualtrics data** section as core and/or helper functions?
2. Every API call is now using `httr::RETRY` and as a result a user may see something like this in the console: 
![image](https://user-images.githubusercontent.com/7210482/149177308-64bb5482-25ed-4f69-836f-33f6c3115028.png) If/when this happens, you should be patient and let it retry multiple times if it's a 50x error. It would probably be helpful to either include some standard paragraph in every function documentation, or put something in the README that covers it.
3. That default message from `httr::RETRY` is vague. If running a call that takes a long time (like `list_distribution_links()` on a large distribution) there may be errors on multiple calls, resulting in something that looks like: 
![image](https://user-images.githubusercontent.com/7210482/149177777-9a8c89ad-1e77-4dcb-ad3a-c44814ea3362.png) It's not readily apparent if this is one issue, and it's retrying it many times, or multiple issues with a single retry. There could be improved messaging to the console to help the user understand what's happening.

juliasilge:
- In the README, we don't list all the other helper functions; it is getting to be enough that I don't think we want to be exhaustive there. However, to help people more, we could add a link to https://docs.ropensci.org/qualtRics/reference/index.html#other-helper-functions
- Let's create a roxygen template to include in the user-facing function documentation about the retrying. [Here are examples of these files](https://github.com/tidymodels/recipes/tree/main/man-roxygen), and we specifically would want to [add one to go to **Details**](https://github.com/tidymodels/recipes/blob/main/man-roxygen/dummy-naming.R).
- I know that the httr2 package lets us have much easier handling of retrying, so we might want to wait to fix that last item. In the meantime we could add more logging only within functions like `list_distribution_links()`, for the paginating maybe?

chrisumphlett:
(1) is already there
(2) would take me a while to understand, I can't take that on right now.
(3) I'm not sure if this is possible. While `res <- qualtrics_api_request("GET", url = fetch_url)` is being evaluated, the errors are caught by `httr::RETRY`. Could something conditionally be written to the console? I could have it write all the pagination to the console but that would be very "chatty". In my experience, with distributions of 25-40K, this tends to be 500-800 pages.

juliasilge:
[...]

PR #251, opened 2022-01-18
Document retrying for user-facing functions

juliasilge:
Addresses #248 

@chrisumphlett can you take a look at this documentation I added and see if you have suggestions based on your experience?


chrisumphlett:
I think that covers the 50x errors well. It would be helpful to also mention that if it's not a 500 error, then retrying is unlikely to help. And then the other api error messages in the package would come in and give more tips based on what the error is.

PR #261, opened 2022-06-08
Add function for fetching in progress responses, closing #220

### LINKED ISSUE #220: Feature: Option to download in progress responses
Would be fantastic to have the option to download in progress as well as completed responses from fetch_survey(). It's helpful for monitoring incoming data before finalising surveys.

jamesmartherus:
Adds a function for downloading in progress responses. Closes #220


R/fetch_inprogress.R
```diff
@@ -0,0 +1,169 @@
...
+#'   unanswer_recode = 999,
+#'   verbose = TRUE,
+#'   # Manually override EndDate to be a character vector
+#'   col_types = readr::cols(EndDate = readr::col_character())
+#' )
+#'
+#' }
+#'
+fetch_inprogress <- function(surveyID,
+                         last_response = deprecated(),
```
juliasilge, line 42: With a new function, we don't need to deprecate any arguments. No one has used it yet, so we can set up a function that only uses current parameters.

R/fetch_inprogress.R
```diff
@@ -0,0 +1,169 @@
...
+#'   label = TRUE,
+#'   unanswer_recode = 999,
+#'   verbose = TRUE,
+#'   # Manually override EndDate to be a character vector
+#'   col_types = readr::cols(EndDate = readr::col_character())
+#' )
+#'
+#' }
+#'
+fetch_inprogress <- function(surveyID,
```
juliasilge, line 41:
```suggestion
fetch_in_progress <- function(surveyID,
```
Let's change the function name so it doesn't squash two words together.

R/fetch_inprogress.R
```diff
@@ -0,0 +1,169 @@
...
+                         import_id = FALSE,
+                         time_zone = NULL,
+                         breakout_sets = TRUE,
+                         add_column_map = TRUE,
+                         add_var_labels = TRUE,
+                         col_types = NULL) {
+
+  if (lifecycle::is_present(last_response)) {
+    lifecycle::deprecate_warn("3.1.2", "fetch_inprogress(last_response = )")
+  }
```
juliasilge, line 64: Same note as above on deprecation

R/fetch_inprogress.R
```diff
@@ -0,0 +1,169 @@
...
+    include_display_order = include_display_order,
+    limit = limit,
+    breakout_sets = breakout_sets,
+    add_column_map = add_column_map,
+    add_var_labels = add_var_labels
+  )
+
+  # See if survey already in tempdir
+  if (!force_request) {
+    if (paste0(surveyID, "_inprogress.rds") %in% list.files(tempdir())) {
```
juliasilge, line 91:
```suggestion
    if (paste0(surveyID, "_in_progress.rds") %in% list.files(tempdir())) {
```

juliasilge: Let's make this change throughout

juliasilge (changes requested):
Thank you so much for this contribution @jamesmartherus! 🙌 

One thing I notice about this PR is that it copies a fair amount of code from `fetch_survey()`. Instead of copy/pasting so much, can we create some helper functions that can be used for both functions?

The other thing we'll need here are some tests. We use vcr for testing, as [outlined in rOpenSci's book on HTTP testing](https://books.ropensci.org/http-testing/).

jamesmartherus:
Thanks @juliasilge! 

I've addressed the deprecation and naming comments and added some tests. Great point about using helpers instead of copying so much code. One potential helper function could check if the survey is already in tempdir?

The reason I ended up copying most of the `fetch_survey()` code is that it already does a really good job using helper functions for checking parameters, generating the URL, building the request, etc. Did you have some other sorts of functions in mind?

jmobrien:
Hey, just seeing this--this is great to have, thanks!  And nice catch on avoiding the saved files overwriting any from `fetch_survey()`

FYI, I just put in a PR adding features to fetch_survey() that also also tried to make the function a bit more modular (and restructured arg checking).  This includes a set of internal helper functions for the export-responses endpoint.  I *think* that could be useful as a drop-in for both `fetch_survey()` and `fetch_inprogress` that could help streamline things w/o too much effort. 

@juliasilge since these two functions are basically the same call, we could in theory consolidate just about everything into a shared internal function except the arg checks (if we're dropping the start/end args), payload/body build (same reason), and the file paths.  But I don't know how far you might want to go on DRY.

juliasilge:
@jamesmartherus Thank you again so much for this new feature! Are you interested in making the changes based on #263? I do think @jmobrien's idea of making one internal function that is used by both of these user-facing functions is the right way to go, for long-term maintainability and avoiding bugs down the road.

jmobrien:
There's a function called `export_responses_request()` that should work for that now.

juliasilge:

Issue #262, opened 2022-06-08
Proposal: get rid of assertthat and replace with unified argument checking/adapting framework w/rlang


jmobrien:
So, I'm trying to finish adding a couple of the really essential features like filtering (#223), time/date control (#158), and survey activation/deactivation (#215, since that's not much extra work once we have date/times going).  

It's all pretty straightforward overall, but I keep running into how complicated our current handling of argument checking is making this.  Right now:

* Argument checks are all done by separate functions made using `assertthat`, (in assertions.R).  
* Generally, each argument (and even each thing to check for each argument), has its own custom-built `assert_*` function.  
* (Most) arguments are actually checked in a separate function (in utils.R) that goes through a list populated with arguments. 
* Those checks work okay, but they're still limited, because some (often optional) arguments can be specified in ways that still need further processing in order to match, say, the API protocol.  
* Any further processing of argument content, if needed, happens somewhere else entirely--for instance inside `construct_raw_payload.R`

`asserthat` is dead--it's not been updated in 5 years.  There are other good options like `checkmate`, but really as some of the recent updates have shown, `rlang::abort()` does quite well.  Also it seems evident that any further development of this kind of thing from the hadley/tidyverse is coming from there (e.g., https://github.com/r-lib/rlang/issues/1111).  

So, maybe continuing with `rlang` might make sense. But more generally while I'm trying to finish all this I was wondering if people wouldn't mind me reworking things more generally into streamlined approach that handles both cleaning and checking arguments in a unified way.  It's probably more efficient for what I'm doing right now and would make everyone's efforts easier going forward.  Thoughts?  If people don't hate the idea I'll send a draft in shortly, and can adjust the strategy from there.

juliasilge:
Oh, I think it would be _GREAT_ if you reworked all that asserting and checking! I think it would be a huge improvement.

jmobrien:
Okay, I'm working on it.  BTW, how do you feel about using the new native pipes (|>)?  It's a lot faster to write with them, but I wanted to run it by you first.

jmobrien:
Glad that it helped.   Closing this, but feel free to reopen if there's more to do.

Issue #215, opened 2021-04-20
New feature request: Deactivate surveys via API


kevintroy:
I have a large number of surveys that I'd like to deactivate without going to the trouble of clicking on each one in the Qualtrics GUI.  I believe it's possible to do this through the [`update survey` endpoint](https://api.qualtrics.com/guides/reference/surveys.json/paths/~1surveys~1%7BsurveyId%7D/put).

(One potential snag: when you deactivate a survey via the GUI, it asks you whether you'd like to close and record responses in progress, or allow those users to finish.  I don't know which of those behaviors we'll get when setting `isActive = FALSE` via the API.)

Is there interest in adding a `deactivate_survey()` function to the package? If so, I could give it a go and file a pull request when I'm done.

juliasilge:
I think that this sounds like good functionality to add to this package, and opening up a PR would be most welcome! 🙌

kevintroy:
I've delved into this just far enough to determine what all needs to be done to make this happen.  Logging here so I can remind myself when I actually write the code.  The approach I'm going to take is to create a generic `update_survey()` function that hits that endpoint, and make `deactivate_survey()` a partial application of that function.

Changes to utils.R:
- Modify `qualtrics_api_request()` to allow `verb = "PUT"` as an option
- Modify `check_params` and `create_raw_payload` to add the parameters used by the `update survey` endpoint (isActive, ownerID, name, and expiration dates)
- Add a generic `update_survey()` function that calls that endpoint

Changes to assertions.R:
Add `isActive` to the parameters checked by `assert_options_logical()`
Add assertions for the other parameters used by the new endpoint

jmobrien:
Turns out the `update survey` endpoint also handles survey renaming, which happens to be something I needed at present for some batch renaming.  So, I've got a working prototype of this that can do both survey renaming and survey activation/deactivation.  

Need to finish this task first but a bit later I will clean it up and make a PR.  

This endpoint also allows pre-specifying start and end times, as well as new survey owners.  I can add the ownership param if desired, but I'll probably hold off on the date params until we work out a general solution for start and end dates in `fetch_survey` as well.

jmobrien:
Just a note - I built on this older endpoint - 
https://api.qualtrics.com/api-reference/reference/surveys.json/paths/~1surveys~1%7BsurveyId%7D/put

It's nice and straightforward for the few things it does.  But then I saw this newer one:

[...]

Issue #158, opened 2020-04-20
ideas: setting up more flexible handling of start and end date params


jmobrien:
Right now, the `start_date` and `end_date` time params for `fetch_survey` just take a simple format of dates per the documentation, but not more precise times.  A more flexible approach to accepting dates/times would help, because with the new V3 fetch endpoint we no longer have the ability to pass a response ID to `last_response` to continue fetching where we left off.  If we could instead pass the last recordedDate entry from a previous fetch, though, we could probably allow something that mimics this functionality.  

However, we'd have to think about flexibility
- if they just give a string date? 
- string date + time?
- What about the format given as output from fetching currently, set by `readr` tools?
- How do we format timeZones, and consider interactions with the API parameter `useLocal`?

Right now, date markers gets implemented in the building of the json payload sent with the fetch request.  Currently accepts only a simply formatted date, not a time, but under the hood adds a blank 00:00:00 time with a "T" to fit the required format: https://api.qualtrics.com/docs/dates-and-times

Played around with some of the base tools when testing #155 to add flexibility, but didn't see what was best.  So far just kept the prior way it was done.  

Options:
- `strptime` and similar was the obvious initial choice.  They do the correct ISO 8601 format, but without the "T" separator.  This can be pretty inflexible, though, as you have to explicitly know the format going in.  
- Also, issues with `strptime` and similar around time zone handling:
   -  wasn't clear to me how to get base tools to implement the time-zone adjustments in a way that was retained in output.
   - The string output for timezones from `readr` (e.g., "CDT" for some data I have in central daylight), doesn't seem to be directly readable by the base tools, as they want the "official" version ("America/Chicago" here)
- still, if we can get this mostly working to get ISO8601, we could probably just do a character replace to add the "T".

- `lubridate` is probably the most obvious next choice, as it has some flexible date parsers and is linked to the tidyverse universe.  But I haven't tested it, esp. whether it does the "T" formatting.  Adds a dependency, of course.  But if it gets close, it might work well.

jmobrien:
Update: `strftime` rather than `strptime` is more versatile and gets closer to what we want, but not quite:

```r 
Sys.time() # Same as style from readr
strftime(Sys.time(), format = "%FT%H:%M:%S%z") # works, except there should be a ":" two spots before the end.
strftime(as.character(Sys.time()), format = "%FT%H:%M:%S%z") # time zone is not included
```
[...]

PR #263, opened 2022-06-14
New argument/error checking approach, new features, addressing #158, #223, & #262

jmobrien:
Parents, talk to your children about the dangers of refactoring code.  

Okay, so this got into a lot, but it's done now, and I think I really managed to clean things up.  Quick list of what's done here:

* All arguments are now handled in a straightforward way with functions in file arguments.R, addressing #262
   * General purpose functions do common checks (string, boolean, integerish, character w/o missing)
   * More detailed checks for specific arguments also perform formatting as needed
   * All error handling operates using rlang::abort (and similar), using cli-type format
   * argument checking/formatting handled directly at the start of each main function
* Removed:
   * assertions.R (and dependency on assertthat, unless I missed something)
   * check_params(), which was redundant and no longer needed
   * Most of construct_raw_payload(), which no longer does argument formatting and just formats request bodies.  This should mean that it's re-usable for other features (e.g., a function for #215)
   * Some old, apparently no longer used code in the main request and response code functions.
* New features:
   * fetch_survey() now fully handles dates & times of Date, POSIX, or equivalent character, closing #158 
       * Updated timezone handling
       * end_date adds 23:59:59 to time when just a date.  This avoids the unintuitive situation where specifying end_date = "2020/03/03" would actually only get cases from 03/02 and before
       * default for timezone is now system locale rather than UTC, which means that things like calling `as.character() on the typical formatting of a datetime object will produce the expected results
    * fetch_survey() now has three `include_*` arguments that allow selection of metadata, questions, and embedded data, closing #223 
    * `fetch_survey()` now uses a more modular structure for handling the multiple calls involved in a fetch.
    * In updating the arg checking and error reporting in `fetch_id()`, added partial name matching & more informative error reporting when multiple surveys are matched.
* Updated tests 
    *  Everything works
    *  Did remove one `webmockr` test from test-qualtrics-api-request.R that didn't seem to work with new format for reasons I couldn't diagnose, but whose functionality seemed to be covered mostly by other tests.


jmobrien:
Commits 
[be6e896](https://github.com/ropensci/qualtRics/pull/263/commits/be6e896fb54ec9ba0ef84e332efe32ee1558383f) through [737f26a](https://github.com/ropensci/qualtRics/pull/263/commits/737f26aa70f846df6b318ce66822b1788ca4f521) are to repair issues with the CMD checks.  (Sorry that took a few more than I intended).

If this works for you, I think we can delete assertthat from Imports

juliasilge (approved):
This is QUITE a lot of work you did here, and what an improvement it will be for the package in the long run!! Thank you 🙌

Issue #253, opened 2022-01-19
Improve the console messaging when there is an API error, now that httr::RETRY is being use


chrisumphlett:
That default message from httr::RETRY is vague. If running a call that takes a long time (like list_distribution_links() on a large distribution) there may be errors on multiple calls, resulting in something that looks like:

![image](https://user-images.githubusercontent.com/7210482/150235396-1d869de5-2031-42ed-a8e2-243419248f27.png)

It's not readily apparent if this is one issue, and it's retrying it many times, or multiple issues with a single retry. There could be improved messaging to the console to help the user understand what's happening.

jmobrien:
@chrisumphlett would it give you what you need if there was a message line showing the survey ID of each request to delineate the groups?

chrisumphlett:
That's not what I had in mind. Many requests will have a single survey id. Really they all do as far as the `qualtRics` function. I personally submit use the `fetch_survey` function on many surveys at once, and call it within a loop where I do print the survey id. But -- the error occurs not on a survey level, but on a page/request level. My example above was calling `list_distribution_links` on a particular distribution used for a single survey.

Knowing the page is not necessarily that useful.  I'm not proposing that the message return _specific_ information. Rather, that it be more descriptive of what is happening so that the user isn't confused (they know that it is retrying the error on purpose, because we believe it was a transient api error/refusal).

jmobrien:
Yeah, that makes sense, and I'm with you.  Do you think these error messages would be beneficial at all, then?  We already have a system for reporting any situation where there's an error that wasn't overcome--i.e., the ones that anyone is likely to care about.

chrisumphlett:
I agree that using the package's standard method is a good idea, however, it gets short-circuited by `httr::RETRY`. 504 errors are handled in `utils.R`:
```
`504` =
          c("Qualtrics API reported a gateway timeout error (504):",
            "These errors are usually resolved by retrying the request",
            glue::glue("instanceId: {httr::content(res)$meta$error$instanceId}"),
            glue::glue("errorCode: {httr::content(res)$meta$error$errorCode}")),
```

But when I get a 504 error, this is what shows, which is what I think will be confusing:
![image](https://user-images.githubusercontent.com/7210482/175771361-180aa10e-cc2b-445c-9c55-37fe026b6527.png)

jmobrien:
[...]

PR #269, opened 2022-06-27
removing redundant warnings from httr::RETRY(), for #253

jmobrien:
Suggested fix for #253 that sets quiet = TRUE in httr::RETRY() so it won't print messages for retries, only printing errors using qualtrics_response_codes() when failed after the 4th attempt.  Also slightly updated the error messages to indicate retries.


juliasilge (approved):
This seems better than the repetitive retry errors. 👍

PR #270, opened 2022-06-27
bugfix for datacenter URL checking

jmobrien:
Small error in how the argument checking handles the removal of trailing slashes in base_url


juliasilge:
Do you think we could add a test for this, maybe in `test-generate-url.R`?

jmobrien:
Sure.  Added to the test-api-credentials.R file since that's where the other tests for checking of base_url were.

juliasilge (approved):
That's great -- thank you so much! 🙌

Issue #264, opened 2022-06-20
what's the verb for something to change survey options?


jmobrien:
Thought I'd hammer out something for #215 and similar, since it shouldn't be too hard with the new framework.   What should we call it, though?  Do we need a new verb?

So far, "fetch" seems generally for downloading content from an endpoint, while I suppose "extract" pulls things out of existing downloads (e.g., `extract_colmap`. )  It's not really an explicit rule so I don't want to overstate the case (`list_distirbution_links()` and `all_surveys()` are both "fetches" of sorts, while `fetch_id() "extracts" something from a dataframe downloaded from `all_surveys()`.   

But if we're about to introduce new things like adjusting options (#215), or adding questions (#228), we're in different territory from anything that came before, so I thought I'd bring the topic up.

jmobrien:
I think it might depend on how further development is to proceed.  One option would be to focus on specialist functions like `deactivate_survey()`, which query the API in a particular way.  Another option would be to focus on functions that generally try to get most/all of the available utility from any given endpoint.

For the latter view, #266 shows 2 new prototype functions that query the *survey-definitions/metadata* endpoint, just in slightly different ways:

https://api.qualtrics.com/a92bfbc7e58fe-get-survey-metadata
https://api.qualtrics.com/ae7f40bbcb91a-update-metadata

These were easy to build--I just added all the available arguments to the body-generating function, then passed it to the querying function.  Really most of the effort was in writing a bit of documentation (only partially done).  

If we go this way, it might make sense to stick with some kind of verb schema for what classes of functions do.  I used our "fetch" and Qualtrics's own "update" for these, but take them just as examples.  (But something like `update_survey_questions()` could work if we stick with that framework, I suppose.)

The alternative would be a specialist route like `activate_survey()` and `deactivate_survey`.  In theory, those could be just as easy to build; just pass only the relevant argument(s). (Though, perhaps a bit harder to maintain).  Under that approach I suppose names would just follow function.  It's really just a design preference b/w a larger number of focused tools vs. a fewer number of consolidated ones matching the API.  Both have advantages I'd think.  What do you prefer?

juliasilge:
I think you're right that specialist functions would be harder to maintain in the long run, and to adjust for various changes in the Qualtrics API. This makes me lean toward some verbs like "fetch" (since it's been in the package already for so long) and "update". 👍

jmobrien:
[...]

PR #275, opened 2022-08-17
add list_directories, list_mailing_lists, and mailing_list_contacts

### LINKED ISSUE #271: Error in `qualtrics_response_codes()` for fetch_distributions() & 0 observations for all_mailinglists()
Dear developers,

I ran into these two issues while running the package, and my API worked well to fetch the survey.

I got 0 observations for all_mailinglists(). No error message.

For fetch-distributions():
```
> distributions <- fetch_distributions(surveys$id[1])
Error in `qualtrics_response_codes()`:
! Qualtrics API raised a bad request (400) error - Please report this on
https://github.com/ropensci/qualtRics/issues
Run `rlang::last_error()` to see where the error occurred.
> rlang::last_error()
<error/rlang_error>
Error in `qualtrics_response_codes()`:
! Qualtrics API raised a bad request (400) error - Please report this on
https://github.com/ropensci/qualtRics/issues
---
Backtrace:
 1. qualtRics::fetch_distributions(surveys$id[1])
 2. qualtRics:::qualtrics_api_request("GET", url = fetch_url)
 3. qualtRics:::qualtrics_response_codes(res)
Run `rlang::last_trace()` to see the full context.
> rlang::last_trace()
<error/rlang_error>
Error in `qualtrics_response_codes()`:
! Qualtrics API raised a bad request (400) error - Please report this on
https://github.com/ropensci/qualtRics/issues
---
Backtrace:
    ▆
 1. └─qualtRics::fetch_distributions(surveys$id[1])
 2.   └─qualtRics:::qualtrics_api_request("GET", url = fetch_url)
 3.     └─qualtRics:::qualtrics_response_codes(res)
 4.       └─rlang::abort("Qualtrics API raised a bad request (400) error - Please report this on\nhttps://github.com/ropensci/qualtRics/issues")
```

Thanks for your help & look forward to getting your reply.

dsen6644:
`list_mailing_lists` will replace `all_mailinglists`, `mailing_list_contacts` will replace `fetch_mailinglist`.

Mailing list pulls now require the directory id associated with the mailing list.

Unit testing is still needed (I'll need some directions, haven't used the VCR package in a while), but I wanted to get other developers to test these new functions first. 

Should close #271

### DIFF
--- a/R/list_directories.R
+++ b/R/list_directories.R
@@ -1,0 +1,40 @@
+#' Retrieve a data frame of all directories from Qualtrics
+#'
+#' @template retry-advice
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' # Register your Qualtrics credentials if you haven't already
+#' qualtrics_api_credentials(
+#'   api_key = "<YOUR-API-KEY>",
+#'   base_url = "<YOUR-BASE-URL>"
+#' )
+#'
+#' # Retrieve a list of all directories
+#' directories <- list_directories()
+#' }
+#'
+
+list_directories <- function(){
+
+  check_credentials()
+
+  # Function-specific API stuff
+  fetch_url <- generate_url(query = "listdirectories")
+
+  elements <- list()
+
+  while(!is.null(fetch_url)){
+
+    res <- qualtrics_api_request("GET", url = fetch_url)
+    elements <- append(elements, res$result$elements)
+    fetch_url <- res$result$nextPage
+
+  }
+
+  x <- purrr::map_df(elements, purrr::flatten)
+
+  return(x)
+
+}
--- a/R/list_mailing_lists.R
+++ b/R/list_mailing_lists.R
@@ -1,0 +1,51 @@
+
+#' Retrieve a data frame of all mailing lists for a directory from Qualtrics
+#'
+#' @param directoryID String. Unique directory ID.
+#'
+#' @template retry-advice
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' # Register your Qualtrics credentials if you haven't already
+#' qualtrics_api_credentials(
+#'   api_key = "<YOUR-API-KEY>",
+#'   base_url = "<YOUR-BASE-URL>"
+#' )
+#'
+#' directories <- list_directories()
+#' mailing_lists <- list_mailing_lists(directories$directoryId[1])
+#'}
+#'
+
+list_mailing_lists <- function(directoryID){
+
+  # qualtrics list mailing lists parameters can be found at
+  # https://api.qualtrics.com/dd83f1535056c-list-mailing-lists
+
+  check_credentials()
+  checkarg_isstring(directoryID)
+
+  fetch_url <- generate_url(query = "listmailinglists",
+                            directoryID = directoryID)
+
+  elements <- list()
+
+  while(!is.null(fetch_url)){
+
+    res <- qualtrics_api_request("GET", url = fetch_url)
+    elements <- append(elements, res$result$elements)
+    fetch_url <- res$result$nextPage
+
+  }
+
+  x <- tibble::tibble(mailingListId = purrr::map_chr(elements, "mailingListId", .default = NA_character_),
+                      name = purrr::map_chr(elements, "name", .default = NA_character_),
+                      ownerId = purrr::map_chr(elements, "ownerId", .default = NA_character_),
+                      lastModifiedDate = purrr::map_chr(elements, "lastModifiedDate", .default = NA_character_),
+                      creationDate = purrr::map_chr(elements, "creationDate", .default = NA_character_),
+                      contactCount = purrr::map_chr(elements, "contactCount", .default = NA_character_))
+  return(x)
+
+}
--- a/R/mailing_list_contacts.R
+++ b/R/mailing_list_contacts.R
@@ -1,0 +1,57 @@
+
+#' Retrieve a data frame of all contacts for a mailing list from Qualtrics
+#'
+#' @param directoryID String. Unique directory ID.
+#' @param mailinglistID String. Unique ID of mailing list you want contacts from.
+#'
+#' @template retry-advice
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' # Register your Qualtrics credentials if you haven't already
+#' qualtrics_api_credentials(
+#'   api_key = "<YOUR-API-KEY>",
+#'   base_url = "<YOUR-BASE-URL>"
+#' )
+#'
+#' directories <- list_directories()
+#' mailing_lists <- list_mailing_lists(directories$directoryId[1])
+#' contacts <- mailing_list_contacts(mailing_lists$mailingListId[1])
+#'}
+#'
+
+mailing_list_contacts <- function(directoryID, mailinglistID){
+
+  # qualtrics mailing lists contacts parameters can be found at
+  # https://api.qualtrics.com/af95194dd116e-list-contacts-in-mailing-list
+
+  check_credentials()
+  checkarg_isstring(directoryID)
+  checkarg_isstring(mailinglistID)
+
+  fetch_url <- generate_url(query = "mailinglistcontacts",
+                            directoryID = directoryID,
+                            mailinglistID = mailinglistID)
+
+  elements <- list()
+
+  while(!is.null(fetch_url)){
+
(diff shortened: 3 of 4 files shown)


jmobrien:
Thanks for this.  

@juliasilge, I will say that this does bring back up the conversation in #264 about whether we want to try and get/keep a standardized verb framing for function names.  

That said, I don't know the mailing list endpoints that well, so @dsen6644 I also think it would be great to have your input on what might work/not make sense tools one would need when working with mailing lists.

dsen6644:
All new functions fall into the "fetch" or "get" category. I decided to use the name of the endpoint in the [api documentation](https://api.qualtrics.com/83b3836dfc351-mailing-list-contacts) for clarity and simplicity, but am completely open to changes to the function names.

juliasilge:
I think in this situation, using the endpoint names makes the most most sense.

@dsen6644 have you gotten anybody who has tried this out yet?

dsen6644:
I have not, @jmobrien and @chrisumphlett could you test these functions and see if they work with your environment?

You should be able to

- Return a df of directory ids using the `list_directories` function.
- Return a df of mailing lists for a directory by passing a directory id to the `list_mailing_lists` function.
- Return a df of contacts by passing a mailing list id and its respective directory id to the `mailing_list_contacts` function.

chrisumphlett:
I used `remotes::install_github("ropensci/qualtRics#275")` to install. I'm not seeing these new functions.

pschatz25:
I'm not seeing the new functions either.

juliasilge:
Apologies! I didn't notice that the new functions weren't exported and documented yet. If you give it a try now via `remotes::install_github("ropensci/qualtRics#275"` it should work now.

chrisumphlett:
It's working for me. I see two potential problems with the data frame returned by `list_mailing_lists()`.
1. datetime columns are in unix epoch which is not consistent with the rest of the package; and, they are not integers (and the decimals are all zeroes so I do not believe it's intended to be partial seconds)
2. `contactCount` column is zero when it shouldn't be.

qual

jmobrien:
To look at this I'd need to set up some fake mailing lists for testing as I don't have any I can touch right now.  

(Does anyone else have a test list on hand?  If so, feel free to drop it here)

jmobrien:
Thanks again @dsen6644! 

Regarding the date/time issue that @chrisumphlett raised, the issue appears to be these `purrr:::map_chr()` calls on lines 46/47:

``` r 
lastModifiedDate = purrr::map_chr(elements, "lastModifiedDate", .default = NA_character_),
creationDate = purrr::map_chr(elements, "creationDate", .default = NA_character_),
```

We usually convert these to POSIXct--but I don't think there's a `map()` (or equivalent) for that output.  So it might be a bit more than a simple fix.  

(That said, IIRC, the nested `map()` calls were something that were used to handle NULLs or missing elements, yes?  You know the raw API output better than I do, @dsen6644--are we likely to encounter these issues here?  If not, would one easy option be to just convert & stack the key variables similarly to how it's done in `all_surveys()?)

`contactCount` should presumably also be numeric (currently character).  But even given that, I don't (yet) see in the code why that would be `0`, unless that's what the endpoint returned.

dsen6644:
we could use `lubridate::as_datetime(purrr::map_dbl(...)/1000))` to solve for the UNIX time.

As for the contact count, it looks like we have to pass a query parameter to the get api request for it to return the actual value... i.e `list(includeCount = "true")`.

@jmobrien what would be the appropriate way to include that in the api request? Right now we have `res <- qualtrics_api_request("GET", url = fetch_url)`.... I tried `    res <- qualtrics_api_request("GET", url = fetch_url, body = list(includeCount = "true")` without success. What's the appropriate way to pass query parameters?

chrisumphlett:
add `query = list(param1= "value1", param2= "value2"))`

dsen6644:
I think I'm actually going to drop the contact count, since it's not "exact"... which could lead to some confusion. See the documentation.

image

jmobrien:
> I think I'm actually going to drop the contact count, since it's not "exact"... which could lead to some confusion. See the documentation.
> 
> image

I think that's probably fine, so long as @juliasilge agrees.  

If it were something to add, something like this would be right as @chrisumphlett mentioned:
> add `query = list(param1= "value1", param2= "value2"))`

except that `qualtrics_api_request()` isn't configured to correctly pass that to *httr* via the `...`.  The ability to add other params might be worthwhile so I'll look into adding that.

dsen6644:
Good to know, it would be important to address if we wanted embedded data associated with our contacts in our mailing lists since we have to pass a similar query `includeEmbedded`.

jmobrien:
Added a `query` arg that should help.  

The more I think about it, maybe keeping `contactCount` would be good?  Like `all_surveys()` the main purpose of 'list_mailing_lists()` is likely getting IDs for use in other functions.  But it's also an at-a-glance summary reference of one's lists--I've certainly used `all_surveys()` that way sometimes.  

Given that, a sense of how many people are on any given list might be useful to users even if it sometimes is imperfect (well, unless it's WAY off, is that likely?).  Think that would be important to note in the function documentation, though.

chrisumphlett:
It's weird that Qualtrics says it's not exact, without a reason. I would
guess that it's probably very close until we see evidence other wise. I
have a lot of lists that I could check it against to validate that
assumption

dsen6644:
`includeCount` (`list_mailing_lists`) and `includeEmbedded` (`mailing_list_contacts`) query parameters have been added.

pschatz25:
At first pass these new functions work well for me. I check the contact counts against my mailing lists in Qualtrics and they match up exactly. 

One other thing I noticed, `all_surveys()` doesn't seem to run now. It just hangs forever.

dsen6644:
I'm seeing performance issues as well... @jmobrien wondering if adding the query parameter has effected anything.

jmobrien:
Hmm, I'll take a look.  In the meantime, do the issues with `all_surveys()` go away (or whatever other performance issues you're having) if you install from the commit just before the new query param?  I guess that would be [f8bc2ea](https://github.com/ropensci/qualtRics/pull/275/commits/f8bc2ea97d36028369e7d142289fab8cb7db6154).  

I think you should be able to do that with `remotes::install_github("ropensci/qualtRics", ref = "f8bc2ea")

pschatz25:
Running  `all_surveys()` with`remotes::install_github("ropensci/qualtRics#275")` did not execute after ~5 minutes. 

Running `all_surveys()` with `remotes::install_github("ropensci/qualtRics", ref = "f8bc2ea")` did execute and returned expected results.

jmobrien:
Thanks for this.  Looks like that's the problem; I'll take a look.

jmobrien:
Finally able to get back to this.  Just made a separate PR #301 for adding queries.  It uses a slightly different  approach that should avoid issues with other types of requests.  

So, this should free things up to proceed here, with one caveat: commit a6f428d is now not needed and should be removed.  @dsen6644, I'm actually not clear how that git procedure is done most appropriately, especially as things center around your fork & branch.  Do you or anyone know?

juliasilge:
Let's give @dsen6644 a few days to catch up and respond if possible, but if he is not available, I think the best/easiest thing to do is a new PR. We can make sure to credit @dsen6644 for his work in the new PR and link to this one.

joshuarbruce:
Hi @juliasilge, @jmobrien and @dsen6644 - just wanted to say thank you to everyone on this thread that helped here, the new functions from #275 are great! Curious if/when they might be merged with the main branch? 

A colleague of mine ran into issues installing the version with `remotes::install_github("ropensci/qualtRics#275")` as referenced above by @pschatz25. However, the solution described installs a slightly different version of the `mailing_list_contacts()` command; it lacks the internal function that pulls embedded data from the referenced mailing list. Installing the package via `remotes::install_github("ropensci/qualtRics#275")` includes the embedded data function. 

Thanks again, this package is super helpful!

benhmin:
Thank you so much for this work - I was having data coercion errors with the fetch_mailinglist() function in version 3.1.7. Seemed to be happening no matter which mailing list ID I used.
`Error: Can't coerce element 1 from a character to a logical`

This version has allowed me to download mailing list contacts using mailing_list_contacts() without errors. Takes a little bit of work to get the directoryID but it's working! Thank you and looking forward to when these are incorporated into the main branch.

juliasilge:

PR #276, opened 2022-08-18
Move @jmobrien to author 🎉

juliasilge:
It has been fantastic to see [all the work](https://github.com/ropensci/qualtRics/commits?author=jmobrien) @jmobrien has contributed here!


jmobrien:
Glad to be a part!  To be sure, I've benefited much more from the package than I've given.

PR #277, opened 2022-08-18
Bugfix & tests for fetch_survey() when include_* = NA, closes #272

### LINKED ISSUE #272: Add either a function or a data.frame in the package to get the list of respondent metadata columns
I would like to be able to have the package provide a global list of the 17 standard metadata columns (plus browser meta info which comes through like response data but should be treated like metadata IMHO), rather than me having them hardcoded. Perhaps others have them hardcoded too; if/when Qualtrics adds another, we'll all need to update our list. If the package holds it then we could automagically have it update whatever we're doing in our code.

Why, you might ask? Here's a [3:42 video with context](https://www.screencast.com/t/B5fPrqe9) on why I have a hardcoded list today, and why you might want to have one as well.

re: the missing metadata I refer to in the video, I was wrong. Qualtrics labels it as "Response Type" but the column is "Status" and that is included.

I can work on this but am curious to hear how those of you who have managed the package long-term feel about including this at all, and if so, whether you'd want it as a function, data.frame, or something else. 

Relevant links from Qualtrics' docs:

* [Browser Meta Info](https://www.qualtrics.com/support/survey-platform/survey-module/editing-questions/question-types-guide/advanced/meta-info-question/)
* [standard metadata on respondents](https://www.qualtrics.com/support/survey-platform/data-and-analysis-module/data/download-data/understanding-your-dataset/#RespondentInformation)

jmobrien:
include_* = NA (which should exclude each class) was getting stripped out of the API request body inappropriately.  Fixed now, with a test added for this.

Addressing #272 - looks like this was the cause of why the thing that @chrisumphlett tried didn't work.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 - Refactored code for checking arguments and errors, thanks to @jmobrien (#263)
 
+- Fixed bug in `fetch_survey()` for `include_* = NA`, thanks to @jmobrien (#277)
+
 # qualtRics 3.1.6
 
 - Add `fetch_distribution_history()` and `list_distribution_links()` for more handling of distribution data, thanks to @chrisumphlett and @dsen6644 (#221, #239)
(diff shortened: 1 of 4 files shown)


juliasilge (approved):

PR #285, opened 2022-09-07
Updated function documentation for fetch_survey()

jmobrien:
@juliasilge feel free to review/edit this. I tried to be detailed to address various known issues/points of confusion, but that does make the text rather long.  I welcome any insight on consolidation (or errors!)

Holding off on rebuilding docs until review, but that can be added before merge.


jmobrien:
Will close #265

juliasilge (approved):
This is definitely long, but I think it's all useful and important information. Let's try this and see if folks find it too difficult to parse (we could edit later and maybe move this to a vignette or similar, if needed).

PR #302, opened 2022-12-07
initial draft of fetch_qsf()

jmobrien:
Example of QSF downloading function, for review.


jmobrien:
@juliasilge new function is for now in same file as its sibling function`fetch_description()`.  Love to hear any further thoughts thoughts about whether it would be better for users to have them together or separate.

juliasilge:
This looks good to me so far! I think I'd probably keep it in the same file as the other, similar function.

jmobrien:
Thoughts on folding it into `fetch_description()` fully?  I lean towards feeling like this works differently enough to warrant being its own function, but I might just not be dialed in to the best arguments for keeping things more compact.

juliasilge:
Yeah, I think another function. What do you think about it being a function that is mainly called for its side effect of writing to disk? Like maybe change the name to `write_psf()` to evoke `write_csv()`, `write_feather()`, and all the functions like that? It could invisibly return the QSF content still, if folks do want to wrangle it in R.

R/fetch_description.R
```diff
@@ -168,3 +171,104 @@ fetch_description <-
...
+#'
+#'
+#'@export
+#'
+fetch_qsf <-
+  function(surveyID,
+           formatting = c("none", "pretty", "compact"),
+           save = FALSE,
+           file = NULL
+  ) {
```
juliasilge, line 210: What if the signature was:

```r
write_qsf <- function (surveyID, file, formatting = c("none", "pretty", "compact") {}
```

jmobrien: Agreed--in the initial design I wasn't focused on saving to disk, but with that focus this this argument order makes more sense.

juliasilge:

jmobrien:
Closing this.  Covered by #319, which uses the `write_qsf()` naming and a more thoughtful implementation.

Issue #304, opened 2023-01-11
Problems with type of `unsubscribed` (logical vs. character)


saudiwin:
An unfortunate new change as of purrr v1.0:

```
In map_chr(), automatic conversion from logical, integer, and double to character is now deprecated. Use an explicit as.character() if needed (#904).
```
This is causing an error in the `fetch_mailinglist` function:

```
<error/purrr_error_indexed>
Error in `purrr::map_lgl()`:
ℹ In index: 1.
Caused by error:
! Can't coerce from a character vector to a logical vector.
---
Backtrace:
 1. qualtRics::fetch_mailinglist(mailId)
 5. purrr::map_lgl(elements, "unsubscribed", .default = NA)
 6. purrr:::map_("logical", .x, .f, ..., .progress = .progress)
 9. purrr:::call_with_cleanup(...)
```

saudiwin:
v0.3.4 or v0.3.5 seem fine, this came in v1.0.0

jmobrien:
Thanks for the input! Yes, this is something we'll need to deal with, likely in the pending functions in #275.

jmobrien:
After looking into this, this appears to be a different issue?    

The problem here is with `map_lgl()` rather than `map_chr()`, which didn't seem to have had the same changes to coercion behavior.  So, the "unsubscribed" column is somehow receiving a character vector.  

And I'm not sure why.  "unsubscribed" is a built-in variable from Qualtrics that should always download as TRUE/FALSE.  I tried to get it to do otherwise but was unsuccessful.  I can generate that error by manually intervening mid-function to change an "unsubscribed" elements to, say, "Y", but I haven't seen it natively.  Is there anything about your lists that might be relevant to this?

Still, what you're describing seems identical to what @benhmin mentioned in #275, so it appears to be possible.  @benhmin did note that those other in-development functions (using the newer API endpoint) seemed to be working.

saudiwin:
Yes, sorry, I was wrong on the error diagnosis (reverted to older version of purrr but same issue). Inspecting the function, yes, for some reason the vector is coming through character:

```
$ unsubscribed         : chr "false"
```

This is inside the `fetch_mailinglist` function. I don't get this error on my Mac laptop; it's in a Digital Ocean droplet I set up. Here is the sessionInfo:

```
R version 4.2.2 Patched (2022-11-10 r83330)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.5 LTS

[...]

PR #305, opened 2023-01-12
patching fetch_mailinglist when "unsubscribed"  is string "true"/"false"

jmobrien:
Addresses #304, and comment in #275

Should smooth over this strange situation (that multiple people have encountered) without compromising existing function.  

Not ideal but should serve until function is replaced with new versions (#275)


jmobrien:
Added second commit fixing issue noticed while working on this.  Entries in mailinglists can vary on what variables they have as embedded data.  If only some cases had data while others had none, though, the mismatch would cause bind_cols() to fail.

juliasilge:
Closed by #306

Issue #212, opened 2021-04-07
`all_surveys` Error When Survey Name is Blank


Scott123456:
When running the `all_surveys` function, I get an error due to a `NULL` survey name. The error occurs when the function tries to `bind_rows(master)` at the end. It cannot complete the `bind_rows` if there is a null value in any of the elements. :/

I was able to work around it by changing the survey name in Qualtrics online. However, it took me a couple hours to figure out the issue. It's kind of crazy that Qualtrics returns an element with a null value (or even allows a survey to have no name), but it does and it breaks the expected behavior of the function.

One proposed solution is to check all elements for null values before appending or binding and replacing them with NA. Though I haven't been able to test this solution yet.

juliasilge:
Oh, that sounds super frustrating, wow! 🥴 Your proposed solution sounds like a good thing to try here in the R package to get around the behavior in the API.

jmobrien:
I noticed that `NULL`'s now no longer cause `bind_rows()` to fail; they just become `NA`'s.  I went looking for this to remember what was up--this is still open, but it looks like it's no longer relevant?  The underlying issue was fixed not long after in dplyr:  https://github.com/tidyverse/dplyr/pull/5502

I tested having a `NULL` survey name, and it's fine now.  Could open up options to simplify other code as well.

PR #306, opened 2023-01-12
New fetch_mailinglist (since now bind_rows takes NULLs)

jmobrien:
Alternative to #305.  With the NULL issues discussed in #212 gone, we have an option to both simplify the code while also fixing the same issues addressed by #304.

This version is more straightforward than the array of `purrr::map`s, and doesn't need the sort of patching they require in #305.  One downside is the code is less explicit about the intended output than before (but there might be a better approach in this vein).

Possibly useful for new upcoming functions `list_mailinglist` and kin.


juliasilge:
I think this option looks quite nice, and like you say, this approach will be extensible. Was it passing the tests locally for you? It seems like we are seeing this error (again?):

> Please use an explicit call to `as.character()` within `map_chr()` instead.

jmobrien:
There are a couple of new functions from purrr/dplyr/tidyr that got introduced here and need to be added to the  roxygen & genereated docs.  Hopefully that will take care of things.

juliasilge (approved):
This is looking great! 🙌

juliasilge:
Yes @pschatz25 for now, you'll need to install from GitHub via `remotes::install_github("ropensci/qualtRics")`.

pschatz25:
Got it. Now I'm back up and running. Thanks, @juliasilge!

Issue #307, opened 2023-02-01
`Error in `export_responses_filedownload()`:` caused by "/" in Qualtrics Project Name


asadow:
I discovered through trial and error that the following error is fixed when a "/" in a project name is removed. 

In my case my survey called "Grounds Shovel Route Afternoon Shift / Weekend Priorities Checklist". Renaming to "Grounds Shovel Route Afternoon Shift - Weekend Priorities Checklist" fixed the error. 

May an error message be implemented that explains the incompatibility of the slash to the user?

`Error in `export_responses_filedownload()`:
! Error extracting CSV from zip file
• The download may have been corrupted; try re-running your query
• Current download file location:
• C:\Users\asadowsk\AppData\Local\Temp\RtmpOSLRWg/temp.zip`

juliasilge:
This makes sense that it would error, because the Qualtrics project name is used as a _filename_ when you download via the API. This is related to #195 and the fix suggested there (using something like `fs::path_sanitize()`) would probably solve both sets of problems and be safer overall.

If someone is interested in submitting a PR implementing a fix in the near term, I will be happy to review! I believe the changes need to be in this function:
https://github.com/ropensci/qualtRics/blob/2d0fb411eb9f0e725983b8b55f1ab5c89055d1b3/R/utils.R#L536
The idea would be to sanitize the path either before or after unzipping. Is it possible to do this without making a copy of the file? Probably.

jmobrien:
Hmm, this might be harder than we think.  As currently configured, the zipped file always has the unproblematic name `temp.zip`.  The problem seems to be with what comes out.  From the api docs:

> The compressed file inside the retrieved ZIP file has this naming convention: {Survey Project Name}.{Export Format}. Note that the file does not return time and date of export. For date and time information, you need to record this information manually during the export process.

There appears to be no way to change that via the API--and there also doesn't appear to be any tools for on-the-fly renaming during extraction using `utils::unzip()`.  So, I'm not sure where we could actually apply a correcting tool like `fs::path_sanitize().`

I suppose it *might* be possible to bypass this? If we use the `list = TRUE` arg from `unzip` we could just get the file name of the internal .csv without attempting to write it to disk.  Then, we could maybe make a direct connection to it with `base::unz()`, then passing it directly `read_csv()` and loading the data into memory.  

I have no idea, though, whether an attempt to link to an internal file like that would still trigger any filesystem naming issues.  I also don't have a Windows box on hand to test this case easily.  

[...]

Issue #195, opened 2020-10-28
fetch_survey Error: File [filename].csv does not exist. Please check if you passed the right file path


stijndelaat:
Thank you for making an R package for Qualtrics. Great idea! I've recently started using it and I am encountering the following error when using the fetch_survey command:

`df <- fetch_survey(surveyID = "SV_xxx", label = TRUE)`
`Error: File [filename].csv does not exist. Please check if you passed the right file path`

The all_surveys() command and the survey_questions() command seem to work fine. Hopefully you are able to help me. Thank you in advance

Rkol8:
Do you have any special characters (":") in the survey title? That was an issue I had and renaming the survey to just characters fixed it.

juliasilge:
Can you run the following code, with your survey ID, and tell me the name of your path? If this is a problem with special characters, we should add a fix to change the file name.

``` r
fetch_url <- qualtRics:::create_fetch_url(Sys.getenv("QUALTRICS_BASE_URL"), "SV_xxxAqx")
raw_payload <- qualtRics:::create_raw_payload(
  label = TRUE,
  start_date = NULL,
  end_date = NULL,
  unanswer_recode = NULL,
  include_display_order = NULL,
  limit = NULL,
  time_zone = NULL,
  include_questions = NULL,
  breakout_sets = TRUE
)

res <- qualtRics:::qualtrics_api_request("POST", url = fetch_url, body = raw_payload)
requestID <- res$result$progressId
survey.fpath <- qualtRics:::download_qualtrics_export(fetch_url, requestID, verbose = TRUE)
#>   |                                                                              |                                                                      |   0%  |                                                                              |======================================================================| 100%
survey.fpath
#> [1] "/var/folders/0w/prb4hnss2gn1p7y34qb2stw00000gp/T//RtmpssoJ6N/Cats Can Have a Little Salami.csv"
```

<sup>Created on 2020-10-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

stijndelaat:
I think it must be a special character issue indeed. I have a ":" in the name of the survey. Also in the link you requested, since that ends with the name ;)

juliasilge:
OK, I am going to leave this open for now, in case we can think about renaming the file. This might not work well, since different OSes have different rules and the problem here is outside of R itself. 

In the meantime, I definitely recommend not using special characters in your survey titles.

dsen6644:
[...]

PR #308, opened 2023-02-08
special chars in files - POC fix for testing

jmobrien:
Alternative approach using `unz` to load files directly, potentially addressing #307 and #195

Needs a lot of cleanup if we use this, but first we should have people (@asadow would you mind?) test it in scenarios that previously produced problems, to see if circumventing writing to files this way represents a viable approach.  

(Works for me for a survey whose filenames include a "/", but that's not definitive as my machine (Mac) already seems to have no issues with those characters.)

### DIFF
--- a/R/read_survey.R
+++ b/R/read_survey.R
@@ -113,6 +113,29 @@ read_survey <- function(file_name,
       header_rows <- 1:2
     }
 
+  # Make connection to zip file: ------
+
+  # Create error handling around unzipping:
+
+  # Unzip and get the filepath for the csv
+  csv_filename_df <-
+    unzip(
+      zipfile = file_name,
+      list = TRUE
+    )
+
+  csv_filename <-
+    csv_filename_df[["Name"]][1]
+
+  cat(paste0("\nInternal filename is: ", csv_filename))
+  cat("\n\n")
+
+  zipcon <-
+    unz(
+      description = file_name,
+      filename = csv_filename,
+      open = "rb"
+    )
 
   # READ RAW DATA ----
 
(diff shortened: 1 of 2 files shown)


juliasilge:
Let me know if I am not understanding this, but do I see that this approach will unzip the zip file twice (once to find out the filename inside and once to unzip to a new filename)? That's probably not too bad but there might be a better approach, like maybe unzipping and then [renaming the file with a sanitized name](https://fs.r-lib.org/reference/path_sanitize.html) or maybe using something from [archive](https://archive.r-lib.org/reference/archive_extract.html) to extract by position. What do you think? Am I maybe not understanding correctly?

Unzipping twice probably won't ever be that slow with files of this size, though, so maybe this option is fine.

jmobrien:
I believe we're not doubling up here. `unzip(list = TRUE)` only (according to the docs) lists the filenames without extraction.   Then `unz()` passes an decompressed binary directly to `read_csv()`.  

So, same number of decompressions, one fewer file write (we only save the .zip, not the .zip and extracted .csv) 

The reason for trying this way is that I'm not sure there even IS a file to sanitize.  Normally, `unzip()` extracts file(s) to disk, then returns a list of extracted file names (here, just one).  The way things are currently set up [here](https://github.com/ropensci/qualtRics/blob/2d0fb411eb9f0e725983b8b55f1ab5c89055d1b3/R/utils.R#L567-L588), we save that name of the extracted file to be passed (eventually) to `read_survey()`.  But, if `unzip()` errors out somehow, we capture it and print the error message seen in #307:

``` r
 # Create error handling around unzipping:
    safeunzip <-
      purrr::possibly(
        utils::unzip,
        NULL
      )


    # Unzip and get the filepath for the csv
    csv_filepath <-
      safeunzip(
        zipfile = tf_path,
        exdir = tempdir()
      )


    if(is.null(csv_filepath)){
      rlang::abort(
        c("Error extracting CSV from zip file",
          "The download may have been corrupted; try re-running your query",
          "Current download file location:",
          tf_path)
      )
    }

```

So, I'm not exactly sure what's happening with @asadow and others (in part because we're masking `unzip()`'s  error message here), but it's at least implied that something is already going wrong: there may in fact be no new extracted file to begin with, and if there is, we don't have its name.  So, what is there to sanitize?  

BTW, @asadow--what happens if you, with a problematically-named survey, just download the usual response .zip  from the Qualtrics web interface, then extract it via the Windows GUI?  What .zip and .csv files do you get then?  

Also, @asadow, might you be willing to test what happens if I remove the error handling from `unzip()` and see what it does directly?

juliasilge:
> So, same number of decompressions, one fewer file write (we only save the .zip, not the .zip and extracted .csv). Plausibly could even be slightly faster, but I haven't tested it.

Ah, you are right, of course! I had missed what `list = TRUE` does in the docs. Given that, I think this PR provides an outline of a good way to go. 👍 Let's try to implement it and see if this can solve some problems.

jmobrien:
#317 aims to handle this now, closing.

Issue #303, opened 2022-12-19
Reorganize .R files around endpoints?


jmobrien:
I've been thinking about the need to revamp URL generation for #294, the number of new and pending new functions, and the casual discussion about where to place new things such as in #299 and/or #302.  Also was doing some things over at *haven* that had me looking at a recent [organization change](https://github.com/tidyverse/haven/commit/b093a6bd20dfee4cf09c3f70a99c8bf80e3803ea).  

Got me wondering if we might do well to start a file structure that organizes functions differently.  Currently wondering about organizing files around major API endpoints, like the ones for [surveys](https://api.qualtrics.com/60d24f6897737-qualtrics-survey-api), [responses](https://api.qualtrics.com/206a07d54ca31-surveys-response-import-export-api), and [mailing lists](https://api.qualtrics.com/0b9b0d24a7a18-mailing-lists).  

Related *potential* ideas, pretty independent so could consider separately:
- organize files by major endpoint groups
- files also have endpoint-associated info/functions that guide url building (since this might help with swapping out to mock servers for tests, and/or can be re-used to build new functions.)
- check functions unique to specific endpoints/functions go in section in same file for easier review/management
- general-purpose check functions folded into utils.R to reduce clutter (delete checks.R)
- others?

juliasilge:
Oh yeah, I think this package could do with some refactoring and reorganization. I personally find it hard to locate something at times! I think `utils.R` has _way_ too much in it, and we should have a file for request building, a file for result parsing, etc.

Like you said, not all of this has to be done at once; some gradual improvements would really get us a better maintenance experience.

jmobrien:
Yeah, I cut down `utils.R` a lot in the last reorganize, and it's still got a lot.  Okay.  I'll poke at it as I finish up some of this other stuff.

PR #317, opened 2023-06-21
remove caching and other persistent writes to disk

### LINKED ISSUE #300: Adjust response caching to consider all differences in API request params?
@juliasilge, thinking about #298--regardless of whether caching is relevant to that issue, caching itself is quite old, and predating a number of query-customization features.  It's now easy to imagine ways that a cached response download could preempt a novel API call to `fetch_survey()`, even though that new call is actually asking for something quite different from what's cached. 

Should this be fixed?  Maybe by changing the IDs for cached files to include other arguments?  Additionally or instead, perhaps by making `force_request = FALSE` the default?

### LINKED ISSUE #314: fetch_survey and save_dir
Hallo, 
I have question. I was looking through the documentation and online vignettes but could not find an answer. 

If I use the fetch_survey function to save my survey locally, is there a way to give the survey a name. With one or two surveys this is not an issue but we have some 30 surveys, thus it would be nice to be able to give it an name when saving it locally. 

mysurvey <- fetch_survey(surveyID = surveys$id[6], 
                         save_dir = "/users/Julia/Desktop/", 
                         verbose = TRUE)

best, 

Andrej

jmobrien:
Primarily does two things:

1. Deprecates response download caching.  Calls to `fetch_survey()` now always query Qualtrics's API, and copies of response downloads will not go to places that are less-than-apparent to users.  Closes #300.  Also closes #314 in the sense that it is no longer relevant (though possibly could still improve vignettes).
2. Restructures file unzipping to better handle special characters in survey names, closing #307 (and superseding #308).  Included here as these changes reduce presence of a different sort of "caching"--i.e. unrequested response data written to disk in temporary folder--which is arguably more responsible.  

Item 2 does not entirely eliminate unrequested disk writes--the .zip file is still saved, albeit hopefully for only a brief time in most cases (see `export_responses_filedownload()`).  AFAIK this is necessary, as I can't find any way to decompress a .zip file held in R's memory as a binary object.  

Also, in implementing number 2, the data-processing activities inside `read_survey()` were moved off to a secondary function `process_raw_survey()`.  Both `fetch_survey()` and `read_survey()` now call `process_raw_survey()` internally (rather than the longstanding call to `read_survey()` by `fetch_survey()`).  Incidental, but probably an organizational improvement on its own.

Finally, helper functions for response exporting were moved out of `utils.R` and into the `fetch_survey.R` in the spirit of #303. Long way to go on that still one still, though. 👍


juliasilge (approved):
This looks great, and will be a real improvement for users @jmobrien! 🙌 

Thanks so much for all this work.

higgi13425:
Here is a survey title (I did not create this one) that still fails to fetch with the dev version - 

Survey of Topics for Clinical Trials Academy 2: Running Your Own Trial

juliasilge:
@higgi13425 Can you open a new issue (rather than add to an older closed PR) with details of your problem?

PR #319, opened 2023-07-12
new function write_qsf(), closes #299, #302

### LINKED ISSUE #299: Automating downloading of complete QSF library from Qualtrics user library
Hi, first of all - thanks for your lovely qualtRics package. It is very useful!

I've encountered a problem, I need to download all my qualtrics surveys - data, metadata and QSF files - from my user account. 

I used qualtRics to automate downloading and saving of both survey data and metadata successfully. However - I have not found a way through qualtRics script to fetch the list, and download QSF files.

Obviously this should not be a problem, as QSF are equivalent to surveyIDs and since identification and opening a connection is already happening through "fetch survey" command. 

Thank you for any help!
Yuval

jmobrien:


juliasilge:
It looks like this is ready for me to try it out, review, merge, etc. Is that right @jmobrien?

jmobrien:
Yep.  Should be ready to go as long as you agree.

R/fetch_description.R
```diff
@@ -168,3 +168,111 @@ fetch_description <-
...
+#'@importFrom stringr str_remove_all
+#'
+#'
+#'@export
+#'
+write_qsf <-
+  function(surveyID,
+           formatting = c("none", "pretty", "compact"),
+           file = NULL,
+           save = TRUE
```
juliasilge, line 206: I think we've got a little bit of [interacting arguments](https://design.tidyverse.org/args-independence.html) here between `file` and `save`. I think the best option is to make `file = NULL` do the same thing as the current `save = FALSE`. That means the default behavior (with `file = NULL`) would be to return the JSON content but not save it, but if someone provides a `file` then it will save it.

R/fetch_description.R
```diff
@@ -168,3 +168,111 @@ fetch_description <-
...
+#'  \code{\link[jsonlite]{prettify}}), or "compact" (more compact, using
+#'  \code{\link[jsonlite]{minify}})
+#'@param file String. Desired filename, including path if desired.
+#'  Typical filename extension is .qsf, though not required (and will not be
+#'  added if absent).  Argument ignored if \code{save == FALSE}.
+#'@param save Boolean. Should downloaded QSF be saved to a file (\code{TRUE},
+#'  default), or returned as output (\code{FALSE})?  Even when \code{TRUE},
+#'  downloaded JSON is returned invisibly.
+#'
+#'
```
juliasilge, line 189: Can we add an example? It would be a nice way to re-emphasize how to save as a file vs. only return the content. It will likely need to be wrapped so it doesn't execute like basically all our other examples.

juliasilge (commented):
This is looking great! 🙌  I've got just a couple of thoughts for changes.

R/fetch_description.R
```diff
@@ -168,3 +168,111 @@ fetch_description <-
...
+    # (Uses "text" plus some string manip below to get at the exact downloaded
+    #  JSON text for QSF.  Bit awkward vs. parsing and reconverting via
+    #  toJSON(), but avoids adding unneeded escape characters.)
+    resp <-
+      qualtrics_api_request(
+        verb = "GET",
+        url = description_url,
+        query = list(format = "qsf"),
+        as = "text",
+        encoding = "UTF-8" # Prevents a warning from guessing encoding
```
juliasilge, line 239: Can we instead make `encoding` an argument to this function so folks can pass in the encoding they need, as outlined in #336? We can still use UTF-8 as the argument default.

jmobrien: See my reply in #336 about this--we can definitely do this if it's valuable, but I think we may have some other things to fix first before being able to properly decide on that.

PR #327, opened 2023-09-01
support for custom temp directories, closes #325

### LINKED ISSUE #325: feature discussion: custom temporary directory
Thinking we should introduce some ability to specify custom temporary directory, and I'm wondering how to implement.  Couple options for this that I'm thinking about, and I'd love your thoughts @juliasilge (on the options and idea in general):

* adding a temporary directory argument to the api credentials function 
* adding in the use of R options (à la https://tibble.tidyverse.org/reference/tibble_options.html)
* adding (yet another!) argument to `fetch_survey()` and other functions for which this might be relevant (either instead of, or in addition to, the first two options.

Context: I've been talking w/some people whose new interest in using `qualtRics` is running up against explicit policies regarding where and how user data can be written to disk.  Now, changes in #317 improved user control on this front for `fetch_survey()` quite a bit; however, because R apparently offers no way to "unzip" a binary representation of a file held in memory, one disk write is still needed to produce a usable dataframe of responses.  While #317 did at least make it so the .zip file would generally only be there for few moments before being deleted, some users may also need to control where that .zip file goes.

jmobrien:
New argument `tmp_dir` (following naming convention in `fs::file_temp()`) for setting location of (brief) zip file write that occurs in `fetch_survey()`.  For when users need particularly tight control over written data location/access.


R/checks.R
```diff
@@ -568,6 +568,20 @@ checkarg_file_name <-
...
   }
 
+#' Check if the temporary directory exists:
+#' @importFrom rlang abort
+#' @keywords internal
+checkarg_tempdir <-
+  function(tmp_dir){
+    if(!dir.exists(tmp_dir)){
+      rlang::abort(
+        c("Error in `tempdir`:",
```
juliasilge, line 578:
```suggestion
        c("Error in `tmp_dir`:",
```
Is this right? We want to use the arg name from `fetch_survey()`?

R/fetch_survey.R
```diff
@@ -67,6 +67,11 @@
 #'   details. Defaults to `NULL`. Overwritten by `convert = TRUE`.
 #' @param verbose Logical. If `TRUE`, verbose messages will be printed to the R
 #'   console. Defaults to `TRUE`.
+#' @param tmp_dir Path to filesystem directory. Qualtrics returns response data
+#'   in compressed (zip) form. To extract raw data, the zip file must be briefly
+#'   written to disk (the file is then promptly deleted). By default, the
+#'   system's temporary directory is used for this (see \code{tempdir()}), but
```
juliasilge, line 73:
```suggestion
#'   system's temporary directory is used for this (see [tempdir()], but
```

Can we link to this?

juliasilge (commented):
This looks great! 🙌  I've got just a couple of questions.

R/checks.R
jmobrien: Yes, my mistake.  The new arg was originally `tempdir`, but that had problems due to overlap with the base function.  Thought I caught all the instances after the change but I missed this one.

R/fetch_survey.R
jmobrien: We should.  And if I hadn't misremembered the proper syntax for how to do that, we already would have 👍.  Appreciate the fix.

Rejecting these specific amendments because I'm in an additional commit that also rebuilds the docs.  Thanks again.

PR #331, opened 2023-11-10
added support for filter_id. closes #224

### LINKED ISSUE #224: Add survey-specific filters to fetch_survey()
Is it possible to add survey-specific filters to fetch_survey()? I can filter after fetching but this would help avoid loading unnecessary info into the environment (helpful when dealing with a huge number of responses). 

For example, I work with a hospital system & often need to get a specific campus' responses. Currently, I can do this with:

fetch_survey(surveyID) %>%
     filter(CAMPUS == campusname)

where CAMPUS is one of the columns from fetch_survey(). It'd be great to apply an arbitrary number of survey-specific filters during the fetch_survey() call.

aaronball85:
This adds a filter_id argument for fetch_survey(). I've updated the code and documentation to reflect the change. Changes passed package tests. Also, I've tested the functionality with a filterId that I created with a Qualtrics survey and it worked as expected. Let me know what you think. Thanks!

### DIFF
--- a/R/fetch_survey.R
+++ b/R/fetch_survey.R
@@ -21,6 +21,9 @@
 #'   `start_date` and/or `expiration_date` when given Date or string objects
 #'   (see above); ignored when these arguments are given POSIXlt/POSIXct
 #'   objects.
+#' @param filter_id String. Unique ID for the response filter created in the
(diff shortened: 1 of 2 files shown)


juliasilge:
Thanks for your patience with us getting to you on this @aaronball85! We will need to get the tests updated in this PR as well. @jmobrien outlined [some of what needs to happen here](https://github.com/ropensci/qualtRics/issues/224#issuecomment-895371031). Would you be able to take a look at this?

Issue #294, opened 2022-11-15
Regenerate all test fixtures


juliasilge:
In ropensci/vcr#255 @sckott suggested it is good practice to regenerate all the test fixtures on occasion. We definitely have not done that, mostly because I (the maintainer) don't have access to many of the features we have tested. These fixtures were created by the individuals who did the original PRs (so involve _their_ surveys).

What is the best way forward from here?

sckott:
vcr does have the `re_record_interval` config option https://docs.ropensci.org/vcr/reference/vcr_configure.html#cassette-options - it is tested and all, but I've not used it  and I don't see anyone else using it in a github search. Other than that, you can just re-record whenever works for you: e.g., whenever making a change to a function which tests use vcr; whenever there's a qualtrics API update - in which case it's a good idea to re-record the cassette to make sure it reflects what you're getting back from the current version of the API.

juliasilge:
Yes, but unfortunately I as a maintainer can't update many (maybe even most?) of the cassettes because I don't have access to those API features.

sckott:
Got it, okay

jmobrien:
I agree that the cassettes for existing tests should probably be re-generated regularly, but I have the same issue with feature availability.  

How does Qualtrics itself view open-source support tools like ours?  Assuming they view our existence at least somewhat positively, perhaps someone there would be open to giving us a full-featured account just for testing?

jmobrien:
Referencing #297, which brought to mind an additional challenge here:

We currently have the situation where all the recorded tests use "www." in the credentials so as to not to be bound to a particular datacenter (and also so it doesn't expose anything about the accounts individual developers rely on for their work).  This actually comes about manually after test generation: the *vcr* fixtures for individual tests are first generated using some specific developer's datacenter ID, then converted after the fact to "www." in both the test R code and fixture YAML.  

Obviously this won't easily work with any kind of regular test regeneration scheme, automated or otherwise.  Realistically we'd probably need something like a package-linked Qualtrics account that contains all the surveys used for tests, or at least someone willing and able to do that with their existing account.

jmobrien:
Looking at the docs, I'm seeing a "mock server" for each endpoint ([example](https://api.qualtrics.com/60d24f6897737-qualtrics-survey-api) ) located at their API dev platform stoplight.io.  Could that be useful here somehow?

juliasilge:
@jmobrien Oh maybe! That would be a really high impact thing to check out.

[...]

PR #343, opened 2024-05-08
Handle ordered factor when multiple choice question uses recoded variable names

### LINKED ISSUE #342: Ordered factor conversion for multiple choice question
It looks like the wrap_mc function in the utils.R script uses "choiceText" from the survey metadata as the factor levels, which isn't guaranteed to match the response values returned from fetchsurveys. This causes an error if, for example, the choice text has a capital letter in it, but the variable name does not. It seems like "variableName" would be a better choice for the levels, though I am new to Qualtrics, so I may be misinterpreting these values. 

https://github.com/ropensci/qualtRics/blob/72bde4e84805d659c98e77c2b92e12a13513081f/R/utils.R#L363

Haunfelder:
If variableNames are non-null, use them for factor levels, otherwise use the choiceText. It may be more appropriate to add an explicit flag to the function to not cause any issues with people relying on choiceText. From my testing, all cases where variableNames do not match choiceText causes an error in fetch_survey as explained in the issue. 

Fixes #342

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # qualtRics (development version)
 
+- Fixed bug when a survey question has *both* recoded values and variable naming thanks to @Haunfelder (#343)
+
 # qualtRics 3.2.0
 
 - Changed how multiple choice questions are mapped to an R factor with the `convert` argument to `fetch_survey()`, to now excluding `NA` as a factor level (#315)
--- a/R/utils.R
+++ b/R/utils.R
@@ -359,8 +359,14 @@ wrapper_mc <- function(data, question_meta) {
 
   # Level names
   ln <- dplyr::pull(dplyr::mutate(meta,
-                                  meta_levels = purrr::map_chr(value,
-                                                               "choiceText")),
+                                  meta_levels = ifelse(
+                                    purrr::map(value, function(x){is.null(x$variableName)}),
+                                    purrr::map_chr(value,
+                                                         "choiceText"),
+                                    purrr::map_chr(value,
+                                                   "variableName")
+                                    )
+                                 ),
                     meta_levels)
   ln <- remove_html(ln)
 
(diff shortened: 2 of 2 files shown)


juliasilge:
@jmobrien do you mind taking a look at this, to see what you think? I don't think I've done much with `variableNames` in my real work.

Haunfelder:
If it helps, here's some more details about the error. I'm fetching a survey that has a multiple choice question with recoding as described in the issue. I use the following code, 

```
test = fetch_survey(surveyID = surveyID, label = T, convert = T,
                    verbose = T, force_request = T)
```
which results in the following warning. 

image

The values of the variable all are set to NA because the factor levels do not match the values returned by the survey API. Here's a picture of the problems output, 

image

The expected value is "Tenure or Tenure-track Professor", but the actual value returned from the API is "tenure or tenure-track professor". 

I think you'll be able to re-create the issue with a single question survey that uses recoding and variable names, but let me know if you need any more details.

jmobrien:
Sure.  I glanced at it earlier--this makes sense and could likely be a good
change, but I should do some testing across a few setups.  I seem to recall
some idiosyncrasies around this part of the metadata() endpoint.  Can't
recall exactly, but I think it might have had something to do with
whether/how it respected the checkbox for these.  So, there are some use
cases we'll want to check first (like if people populate these fields, but
then turn them off)

juliasilge (approved):
Thank you so much for this PR @Haunfelder, and for your patience! 🙌 

It would be nice to get this behavior under testing, but we are looking at totally revamping our testing strategy as outlined in https://github.com/ropensci/qualtRics/issues/294#issuecomment-1345371583 so I don't think it's a high priority for now.

Haunfelder:
Awesome, thanks @juliasilge! I appreciate you all taking the time to understand the issue I was seeing. The qualtRics package has been very useful for me!

Issue #345, opened 2024-06-17
Error in unz (colon in survey title)


higgi13425:
Apprently unable to connect/unzip a survey file with a colon in the title.
The survey is entitled 

Survey of Topics for Clinical Trials Academy 2: Running Your Own Trial

Code is 
my_survey <- fetch_survey(surveyID = 'SV_9Kw9QbuMrMyOoUC')
  |===============================================| 100%
Error in unz(description = zip_path, filename = csv_filename, open = "rb") : 
  cannot open the connection
In addition: Warning message:
In unz(description = zip_path, filename = csv_filename, open = "rb") :
  cannot open zip file '/var/folders/93/s18zkv2d4f556fxbjvb8yglc0000gp/T/RtmpAYTrax/file1835139200ab0.zip:Survey of Topics for Clinical Trials Academy 2'

This appears to cut off the path at the colon, so that it is an incorrect path.

Ideas on how to fix?

juliasilge:
Ah yep, this is related to #329 and also #195. Since we have to deal with file paths on disk here in this R package and unzipping files, we can't handle special characters like `:` in survey titles. Eventually we would like to provide a way for you to change your survey title from R, but for now the workaround is for you to change your survey title in the Qualtrics UI. I know that is likely disappointing! 😞

gdmcdonald:
Same issue, but the workaround of renaming the survey works fine. 👍

higgi13425:
Originator of survey left employment - finally got someone in IT to change the title.
This did work

Issue #329, opened 2023-09-28
fetch_survey error; filename issue?


BeardedJayhawk:
OK, when I attempt to run fetch_survey I get an error. This is the simplest version of the code that's causing issues:

`mysurvey <- fetch_survey(surveyID = 'SV_0UFBbXXXXXXXXX')`

and this is the error:

```
Error in `checkarg_file_name()`:
! Error in `file_name`:
* The file given does not exist: C:/Users/<username>/AppData/Local/Temp/RtmpuUO8XD/Survey Name: The Company Survey - 2023-03 March Form.csv
Run `rlang::last_trace()` to see where the error occurred.
> rlang::last_trace(drop = FALSE)
<error/rlang_error>
Error in `checkarg_file_name()`:
! Error in `file_name`:
* The file given does not exist: C:/Users/<username>/AppData/Local/Temp/RtmpuUO8XD/Survey Name: The Company Survey - 2023-03 March Form.csv
---
Backtrace:
    x
 1. \-qualtRics::fetch_survey(surveyID = "SV_0UFBbXXXXXXXXX")
 2.   \-qualtRics::read_survey(...)
 3.     \-qualtRics:::checkarg_file_name(file_name)
 4.       \-rlang::abort(c("Error in `file_name`:", glue::glue("The file given does not exist: {file_name}")))
```

I of course edited out/changed some details, but one I did not change is the colon in the survey name. when I look at the temp directory there is a temp file named just "Survey Name".

Is the colon somehow causing the issue? This isn't happening with other surveys that don't have a colon in the name.

juliasilge:
Yes, that's right. You can read more in #195 about the kinds of problems that people run into with special characters in their survey titles, which translates to special characters in a _filename_ which of course is no good. Since #195 was opened, we have taken on a fs dependency so we could consider using `fs::path_sanitize()` to try to fix these up.

juliasilge:
Oh wait, `fetch_survey()` doesn't call `read_survey()` anymore! Can you try installing from GitHub and seeing if the development version works for you?

```r
remotes::install_github("ropensci/qualtRics")
```

BeardedJayhawk:
alright, now I'm getting a different error:

```
Error in unz(description = zip_path, filename = csv_filename, open = "rb") : 
  cannot open the connection
In addition: Warning message:
In unz(description = zip_path, filename = csv_filename, open = "rb") :
[...]

PR #348, opened 2024-07-17
use system unzip if colon in title #345 #329 (draft)

WillForan:
fetch_survey:export_responses_filedownload can use external unzip binary This streams zip file contents into a temporary csv file to avoid unz()

TODO: check system('which unzip') or error?

### DIFF
--- a/R/fetch_survey.R
+++ b/R/fetch_survey.R
@@ -525,13 +525,23 @@ export_responses_filedownload <-
     }
 
 
+    if(grepl(":", csv_filename)){
+       warning("Survey title cannot be a file name '",csv_filename, "'")
+       #if(grepl("'|\\\\",zip_path) ) stop("unsafe zip_path ", zip_path)
+       tmp_csv_path <- fs::file_temp(ext = "csv", tmp_dir = tmp_dir)
+       withr::defer(fs::file_delete(tmp_csv_path))
+       unsafe_unzip_cmd <- paste0("unzip -p '",zip_path,"' > '",tmp_csv_path,"'")
+       system(unsafe_unzip_cmd)
+       zipcon <- tmp_csv_path
+    } else {
     # Make connection to zip file:
-    zipcon <-
-      unz(
-        description = zip_path,
-        filename = csv_filename,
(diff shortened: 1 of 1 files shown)


juliasilge:
This unfortunately is not going to work on Windows AFAIK, so we do need to find a solution for #345, #329, and friends that does not rely on a `system()` call. 😕

Thank you so much for your interest in contributing, though!

WillForan:
it looks like windows includes [`tar.exe`](https://ss64.com/nt/tar.html) which can [unzip .zip](https://www.liberiangeek.net/2024/03/zip-unzip-files-windows-command-line/) to stdout. I think that looks like `tar.exe -Oxf zipfile.zip`. Though I'm not sure what file redirect looks like with R's system() on windows. maybe `unizip -p survey.zip > tmp`  could be `tar -Oxf survey.zip > tmp` when `grepl("windows", .Platform$OS.type)`? If this kludge is even worth pursuing

WillForan:
alternatively, instead of creating a temp file, maybe the tar.exe/unzip to stdin could be captured with `pipe()`?


quick test on linux looks promissing
```
> system("echo 'a,b,c' > test.csv; zip test.zip test.csv")
  adding: test.csv (stored 0%)
> read.csv(pipe('unzip -p test.zip'),header=F)
  V1 V2 V3
1  a  b  c
```

juliasilge:
Thank you so much for your thoughts @WillForan! I don't know that using a `system()` call is a very robust way to go in general, but I did open up #349 that uses the [archive](https://archive.r-lib.org/) package instead of `unz()` from base R, because it has better file path handling. I think between using archive and fs, we may have solved the problem.

Want to take it for a spin, with `devtools::install_github("ropensci/qualtRics@use-archive-extract")`?

WillForan:
Thanks! Works for me! I just needed an extra `apt install libarchive-dev` -- and `install.packages('archive')`'s failure had a useful message instructing me to do so. 

Closing this PR

juliasilge:
That is great news! Thank you for trying it out @WillForan. 🙏

PR #355, opened 2024-09-04
Extract whole archive, to avoid problems with special characters

### LINKED ISSUE #354: fetch_survey tmpdir problem in 3.2.1?
Package version 3.2.0 works perfectly for retrieving data, but 3.2.1 fails with identical code with this error message:

> Error: '/var/folders/yf/q87d8g990l5f_sz039gjgmy80000gn/T/RtmpTiIVMj/name of my Qualtrics project.csv' does not exist.

I have tried to change the tmpdir, but get the same error. Downgraded to 3.2.0 and things work again, in the same R session.

juliasilge:
Closes #354 

It turns out that if we try to do:

```r
archive::archive_extract(zip_path, tmp_dir, csv_filename)
```

with a `csv_filename` with special characters in the name, nothing happens; somehow `archive_extract` does not find the file. This may be related to https://github.com/r-lib/archive/issues/75.

In our case, I don't think there is any harm in just extracting the whole archive; it is in the temp directory and I've never seen anything in the zip archives from Qualtrics but the CSV file themselves. With the change in this PR, the CSV does get extracted and then I can access it via `fs::path(tmp_dir, csv_filename)`.

### DIFF
--- a/R/fetch_survey.R
+++ b/R/fetch_survey.R
@@ -523,7 +523,7 @@ export_responses_filedownload <-
     }
 
     # Extract CSV from zip file:
-    archive::archive_extract(zip_path, tmp_dir, csv_filename)
+    archive::archive_extract(zip_path, tmp_dir)
 
     # Read in raw data:
     rawdata <-


juliasilge:
This PR can be installed via `devtools::install_github("ropensci/qualtRics@just-extract-whole-archive")` to test it out.

@pgmj would you be willing to try it out?

pgmj:
Sorry for the delay in responding. I tried it and it works! Thanks a lot for fixing this, and for making this super useful package!

PR #360, opened 2024-12-22
Replace `surveys$id[i]` with `"SV_xxxxxxx1"` to close #341

### LINKED ISSUE #341: Question/ comment about the documentation
Hallo, 
I very much like your package!

I have a question. Not about a bug but a comment on the documentation. 

I was wondering why all of the examples that are used in the documentation use the location: my_survey <- fetch_survey(surveyID = surveys$id[6]) instead of the  string and the Unique ID. Like this: fetch_survey(surveyID = "SV_brr7rymdAnaGekS", 

In the documentation, the use of the string and the Unique ID is explained: 

surveyID
String. Unique ID for the survey you want to download. Returned as id by the [all_surveys](https://docs.ropensci.org/qualtRics/reference/all_surveys.html) function.

However, all of the examples use the surveys$id[6]. 

I would imagine that using the unique id (fetch_survey(surveyID = "SV_brr7rymdAnaGekS", ) would be preferred to the location, due to precision and due to consistency over time. 

Or am I missing something?

JeffreyRStevens:
This PR addresses #341 by replacing `surveys$id[i]` with `"SV_xxxxxxx1"`  in all documentation. This seems like a more robust way to name the survey ID, since deleting surveys could change a particular survey's position in the list of surveys. Therefore, I made the replacement in all instances in the README, vignette, and examples but describe how to use `surveys$id[i]` in the vignette. But I'm happy to alter this, if desired.

### DIFF
--- a/R/all_surveys.R
+++ b/R/all_surveys.R
@@ -17,10 +17,10 @@
 #' surveys <- all_surveys()
 #'
 #' # Retrieve a single survey
-#' mysurvey <- fetch_survey(surveyID = surveys$id[6])
+#' mysurvey <- fetch_survey(surveyID = "SV_xxxxxxx1")
 #'
 #' mysurvey <- fetch_survey(
-#'   surveyID = surveys$id[6],
+#'   surveyID = "SV_xxxxxxx1",
 #'   save_dir = tempdir(),
 #'   start_date = "2018-01-01",
 #'   end_date = "2018-01-31",
--- a/R/column_map.R
+++ b/R/column_map.R
@@ -19,11 +19,11 @@
 #' surveys <- all_surveys()
 #'
 #' # Retrieve column mapping for a survey
-#' mapping <- column_map(surveyID = surveys$id[6])
+#' mapping <- column_map(surveyID = "SV_xxxxxxx1")
 #'
 #' # Retrieve a single survey, filtering for specific questions
 #' mysurvey <- fetch_survey(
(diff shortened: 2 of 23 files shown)


juliasilge:
Thank you so much for this PR, @JeffreyRStevens!

I don't want to change all these use cases like this based on other feedback and use cases for our docs, but we can definitely _add_ it to the docs for `fetch_survey()` (not replace the current approach). For example, we could edit those exmaples to something like:

```r
#' # Retrieve a list of surveys
#' surveys <- all_surveys()
#'
#' # Retrieve a single survey using your `surveys` dataframe:
#' my_survey <- fetch_survey(surveyID = surveys$id[6])
#'
#' # Retrieve a single survey if you know the survey ID as a string:
#' my_survey <- fetch_survey(surveyID = "SV_xxxxxxxxxxxxxxx")
#'
#' # Retrieve your survey specifying other args:
#' my_survey <- fetch_survey(
#'   surveyID = surveys$id[6],
#'   start_date = "2018-01-01",
#'   end_date = "2018-01-31",
#'   limit = 100,
#'   label = TRUE,
#'   unanswer_recode = 999,
#'   verbose = TRUE,
#'   # Manually override EndDate to be a character vector
#'   col_types = readr::cols(EndDate = readr::col_character())
#' )
```

Are you interested in updating this PR, or doing a separate one for this change?

PR #372, opened 2025-07-23
Attempt some re-recording of fixtures

juliasilge:
Addresses #294 with some of the features that I do have access to


juliasilge:
I'm trying again in #374

Issue #368, opened 2025-07-16
Fixes for dev version of vcr


sckott:
Hey, we're releasing a new version of vcr very soon `pak::pak("ropensci/vcr")`. 

Running your tests shows that you'll need to re-record your cassettes with the new version of vcr, should be easy to do, just delete your current cassettes and run your tests 2x to make sure they play back correctly as well. 

In addition, we've moved default cassette storage location to `testthat/_vcr` alongside `_snaps` - that's up to you of course where cassettes go. 

I see some tests failing that use cassettes as well with dev vcr - we have a new logging tool to help you debug - see `local_vcr_configure_log()` to get more details

https://docs.ropensci.org/vcr/

juliasilge:
Ah, unfortunately, we as the maintainers cannot re-record our cassettes as we outline in #294. What would you recommend we do in that situation?

Maybe what we need to do is remove the mocked tests altogether and go with only doing testing via the mock servers as I started in #350 and #351.

juliasilge:
I could also use some docs on how to re-record cassettes, for the ones where I do have access to the API feature. I looked at:

- https://docs.ropensci.org/vcr/
- https://books.ropensci.org/http-testing/

But I could not find how to update or re-record an existing cassette.

sckott:
> I could not find how to update or re-record an existing cassette.

We don't have it, so it's not you. 

The brief answer is: delete the cassette you want to re-record, then run the test again that makes that cassette. 

The longer answer is:
- There's no functions in vcr for help with re-recording but there is https://docs.ropensci.org/vcr/reference/vcr_configure.html#arg-re-record-interval arg in configure
- There's the config option `re_record_interval` - see https://github.com/ropensci/vcr/blob/main/R/cassette_class.R#L513-L548 - You could do this locally or have a GH action that sometimes records new cassettes depending on how long ago the cassettes were recorded. 
- I've done a cron job on gh actions to rewrite cassettes on a schedule (1/week) in some Python projects that use the main python vcr pkg - and the rewrites open up a PR for a human to review - but one could have the changes committed directly to main too

Open to suggestions on what would be intuitive to you on this front

sckott:
[...]

PR #373, opened 2025-07-23
Update metadata.yml fixture for new vcr version

sckott:
Did this by hand. This should be what vcr v2 would record, but I can't be 100% certain 

I'll get to the others tomorrow

#368


juliasilge:
Thank you for working on this! I did figure out some of my problems related to headers that were being recorded in ways that we were not expecting. FWIW I was never able to get the cassettes working again on CI for these sets of tests:

- https://github.com/ropensci/qualtRics/blob/main/tests/testthat/test-fetch-survey.R
- https://github.com/ropensci/qualtRics/blob/main/tests/testthat/test-extract-colmap.R

Those all involve tests that run `fetch_survey()` (the main function of the package) and I was not able to generate a cassette that could be run in CI. I am running those tests live now, after adding my real credentials to CI here, and went ahead and removed the cassettes.

sckott:
Okay!

Issue #386, opened 2026-04-29
Migrate `all_mailinglists()` and `fetch_mailinglist()` to new Qualtrics endpoints


ProfFancyPants:
I want to see if anyone had incites regarding this email I received from Qualtrics, and whether it impacts how qualtrRics's functions make their API calls or it is something I'm doing specifically. It seems to do with the calls mainly to contact lists, and segments. The email looks as follows:

# Qualtrics API Deprecation Notice

You're listed as a Brand Administrator for XXX, which has active usage of one or more of the endpoints being removed. Several XM Directory Public API endpoints will be fully deprecated by June 30, 2026. Here's what you need to know to avoid disruptions.

## What's Being Removed?

All legacy _Research Core Contacts_ API endpoints will be removed.
A full list of replacements is available in the migration guide:
👉 https://api.qualtrics.com/c4c0dbec83622-research-core-to-xm-directory-migration-guide

In addition, the following XM Directory endpoints are being deprecated:

## Contact Frequency Rules
[Get Contact Frequency Rule](https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fapi.qualtrics.com%2Fe0d9ace669f80-get-contact-frequency-rule&data=05%7C02%7CRBruner%40highscope.org%7C00b02ea9a689413ac63c08dea61b53e9%7C4c7e43bff74640a78194e1f6a5e2a4ff%7C0%7C0%7C639130833480161387%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=o1J9XFhAweDOOcAdBGiJ4wdZ88pN88R06qOe1U7cQrU%3D&reserved=0)
[...]

PR #388, opened 2026-06-15
Migrate mailing list functions to XM Directory API

apilipshen:
## Summary
  
  - `all_mailinglists()` and `fetch_mailinglist()` previously used the legacy Research Core Contacts endpoints which Qualtrics is retiring on June 30, 2026 (#386)
  - Migrates both functions to the XM Directory API endpoints (`/directories/{directoryId}/mailinglists`)
  - Directory ID is discovered automatically from the API and cached for the session, so existing user code requires no changes
  - Updated tests to reflect the new XM Directory response schema
  - Verified working against a live Qualtrics account

  ## Test plan

  - [ ] Unit tests pass (`test-all-mailinglists.R`, `test-fetch-mailinglist.R`)
  - [ ] Live test: `all_mailinglists()` returns expected results
  - [ ] Live test: `fetch_mailinglist()` returns contacts for a given list


apilipshen:
The Windows CI failure is a 429 rate-limit error from the Qualtrics API on pre-existing fetch_survey() tests — not related to this PR's changes.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # qualtRics (development version)
 
+- Migrated `all_mailinglists()` and `fetch_mailinglist()` from the deprecated
```
juliasilge, line 3: The new endpoints return a different shape, which you can see from the updated tests:
- `all_mailinglists()`: `id` ➡️  `mailingListId`; `libraryId`/`category`/`folder` dropped; `contactCount`/`ownerId`/dates added.
- `fetch_mailinglist()`: `id` ➡️  `contactId`; `externalDataReference` ➡️  `extRef`.

Existing code that references `x$id` or `x$externalDataReference` will silently get `NULL`. The NEWS line here "existing code requires no changes" is accurate only about directory discovery, but readers will probably take it as full backward compatibility. Could you update NEWS to call out the renamed and changed columns explicitly?

R/fetch_mailinglist.R
juliasilge, line 1: Can you update the roxygen docs in this file?

- The `@param mailinglistID` description says it is "Returned as `id` by all_mailinglists"; that field is now `mailingListId`.
- The example `fetch_mailinglist(mailinglists$id[1])` now passes `NULL`. It should be `mailinglists$mailingListId[1]`.

R/utils.R
```diff
@@ -135,36 +135,58 @@ generate_url <-
...
   glue::glue("https://{base_url}/API/v3")
 }
 
+#' Retrieve and cache the user's default XM Directory ID
+#'
+#' Checks QUALTRICS_DIRECTORY_ID env var first; if unset, calls GET /directories
+#' to discover it and caches the result for the session.
+#'
+#' @keywords internal
+fetch_directory_id <- function() {
```
juliasilge, line 175: IIUC `fetch_directory_id()` silently picks the first directory here, and also IIUC accounts can have more than one XM Directory (although I'm not an expert on this). If that's right, `dirs[[1]]` could quietly target the wrong one with no signal to the user. The `QUALTRICS_DIRECTORY_ID` escape hatch is really nice, but it is not documented anywhere user-facing. Could we document that env var in the user-facing function (not the internal one), and warn (is error too much?) when multiple directories are returned so users know to set it?

tests/testthat/test-all-mailinglists.R
```diff
@@ -1,17 +1,49 @@
...
+      "https://stoplight.io/mocks/qualtricsv2/publicapidocs/60928",
+    paginate_api_request = function(fetch_url) {
+      list(list(
+        contactCount     = 5L,
+        mailingListId    = "CG_abc123",
+        name             = "Test List",
+        lastModifiedDate = 1700000000000,
+        creationDate     = 1700000000000,
+        ownerId          = "GR_abc123"
+      ))
```
juliasilge, line 20: These tests now mock `paginate_api_request` with hand-crafted data rather than the stoplight mock. Is this because of network issues? You confirmed these against a live response?

juliasilge (commented):
Thanks for taking this on so much! 🙌  I'm really glad that we might be able to sneak this XM Directory migration in ahead of or very near the June 30, 2026 retirement.

All of this is really great:

- the automatic directory discovery with a `QUALTRICS_DIRECTORY_ID` override
- session caching that follows the package's existing pattern (cleared in `zzz.R` alongside the API key and base URL)
- really nice `generate_url` refactor

Tests pass locally for me, and the Windows CI passed when I clicked to re-run it. I do notice that both tests set `QUALTRICS_DIRECTORY_ID`, which short-circuits `fetch_directory_id()` before it calls the API, so the discovery branch and the "No XM Directories found" abort has no coverage right now. What do you think about a small test that mocks `qualtrics_api_request` to exercise both paths?

NEWS.md
apilipshen: Good call — updated. I reworded the "no changes" line so it's clearly scoped to the  directory discovery, and added explicit bullets listing the renamed/dropped/added columns for both functions (`id` → `mailingListId`, `externalDataReference` → `extRef`, etc.). Commit 0139d04.

R/fetch_mailinglist.R
apilipshen: Fixed — `@param mailinglistID` now says `mailingListId`, and the example uses `mailinglists$mailingListId[1]`. Commit 0139d04.

R/utils.R
apilipshen: You're right that an account can have more than one directory.  `fetch_directory_id()` now emits a warning when it finds multiple, naming the one it picked and pointing to `QUALTRICS_DIRECTORY_ID`. I went with a warning rather than an error so existing single-directory setups aren't interrupted — happy to make it an abort if you'd prefer. I also documented the env var in the `@details` of both user-facing functions. Commit 0139d04.

tests/testthat/test-all-mailinglists.R
apilipshen: Good question. These two tests stub `paginate_api_request` with inline data rather  than going through the stoplight mock, because the mock doesn't have the new  directory-scoped route. Hitting `/directories/{id}/mailinglists` on the mock returns Prism's "Route not resolved, no path matched" (as a 422). The only route
that still resolves is the old flat `/mailinglists`, and it returns the deprecated Research Core schema  (`libraryId`/`id`/`category`/`folder`) — i.e. exactly what we're migrating away from. So there's no way to exercise the new endpoints through that mock; stubbing `paginate_api_request` also drops the network dependency.

The stub data isn't guessed, though — I confirmed the response shape against a live call to my own Qualtrics account (214 mailing lists and 509 contacts) and copied the actual field names and types into the tests. That live run is where the schema changes (`id` → `mailingListId`, `externalDataReference` → `extRef`, etc.) surfaced.

I kept these as inline stubs rather than a recorded `vcr` cassette on purpose: `fetch_mailinglist()` returns live contact records (my test call pulled 509 real contacts), and I didn't want real names/emails landing in a cassette in a public repo. If you'd prefer a cassette, I'm happy to record one against a small throwaway mailing list with synthetic contacts so there's nothing sensitive to scrub — just say the word.

apilipshen:
On the test coverage — added a dedicated `test-fetch-directory-id.R` that leaves `QUALTRICS_DIRECTORY_ID` unset and mocks `qualtrics_api_request`, covering the discovery path, the "No XM Directories found" abort, and the new multiple-directory warning. Commit 0139d04.

tests/testthat/test-all-mailinglists.R
juliasilge: That makes a lot of sense to me; no need to change!

R/utils.R
juliasilge: That's a good argument for a warning; makes sense!

juliasilge (approved):
Thank you so much for working on this!  🚀

